PR #131, opened 2025-04-20
UI console dev

jnrjose:


jerry-enebeli:
Hi @jnrjose, thanks for the PR. Unfortunately, I’ll have to decline it. Blnk Core does not support a UI, and we don’t plan to support an official UI in this repo in the future either. Our UI is currently part of our cloud solution.

PR #133, opened 2025-04-29
feat: Add Kubernetes manifests for deployment and management

brenmat:
This commit introduces Kubernetes (k8s) manifests to support the deployment, scaling, and management of the Blnk core.

### DIFF
--- a/infrastructure/k8s-manifests/postgres-statefulset.yaml
+++ b/infrastructure/k8s-manifests/postgres-statefulset.yaml
@@ -53,8 +53,12 @@ spec:
               name: pg-data
               subPath: pgdata
       restartPolicy: Always
-      volumes:
-        - name: pg-data
-          persistentVolumeClaim:
-            claimName: pg-data
+  volumeClaimTemplates:
+    - metadata:
+        name: pg-data
+      spec:
+        accessModes: ["ReadWriteOnce"]
+        resources:
+          requests:
+            storage: 10Gi
 status: {} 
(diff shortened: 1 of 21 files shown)


infrastructure/k8s-manifests/postgres-pod.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: postgres
+  labels:
+    app: postgres
```
jerry-enebeli, line 6: I think the label should be 

```
labels:
    io.kompose.service: postgres 
```

infrastructure/k8s-manifests/postgres-statefulset.yaml
```diff
@@ -0,0 +1,60 @@
...
+          name: postgres
+          ports:
+            - containerPort: 5432
+          resources: {}
+          volumeMounts:
+            - mountPath: /var/lib/postgresql/data
+              name: pg-data
+              subPath: pgdata
+      restartPolicy: Always
+      volumes:
```
jerry-enebeli, line 56:
```suggestion
  volumeClaimTemplates:
    - metadata:
        name: pg-data
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 10Gi
```

I think using a volumeClaimTemplates might be better for a statefulset, this way each pods gets its own dedicated storage volume.

jerry-enebeli (changes requested):

jerry-enebeli (approved):

PR #189, opened 2025-09-08
Redis ping check

### LINKED ISSUE #188: blnk can start without redis
Redis is required for blnk, but it's possible to start server without it.

**To Reproduce**
1. start blnk without redids
2. create transaction
3. receive 400 status code with message cannot dial redis

**Expected behavior**
blnk shouldn't start without redis and status code shoudn't be 400

**Screenshots**

Image

RyanCarlisle:
Fixes #188 

tr1

### DIFF
--- a/internal/redis-db/redis_test.go
+++ b/internal/redis-db/redis_test.go
@@ -71,11 +71,16 @@ func TestNewRedisClient(t *testing.T) {
 		{
 			name:      "single address",
 			addresses: []string{"localhost:6379"},
-			wantErr:   true,
+			wantErr:   false,
 		},
 		{
 			name:      "multiple addresses",
 			addresses: []string{"localhost:6379", "localhost:6380"},
+			wantErr:   false,
+		},
+		{
+			name:      "single address with wrong port",
+			addresses: []string{"localhost:65749"},
 			wantErr:   true,
 		},
 	}
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -54,6 +54,7 @@ services:
   redis:
     image: redis:7.2.4
     container_name: redis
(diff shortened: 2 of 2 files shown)


internal/redis-db/redis_test.go
```diff
@@ -71,12 +71,12 @@ func TestNewRedisClient(t *testing.T) {
 		{
 			name:      "single address",
 			addresses: []string{"localhost:6379"},
-			wantErr:   false,
```
jerry-enebeli, line 74: I think `wantErr` should be `false`, given that the URL is correct, present, and uses a standalone(single URL).

It'd only be `true` when the Redis instance is not running. 

We can extend the test to add a wrong port or host and expect an error(the ping should throw an error)

jerry-enebeli (changes requested):
Thank you for the PR @RyanCarlisle

internal/redis-db/redis_test.go
RyanCarlisle: The issue here is if Redis is not set up already, and you run the test case, the test case itself will panic as it will be trying to assert on NoError. This is the reason why I had made wantErr `true`. Could you let me know what is to be done instead in such cases?

jerry-enebeli:
```suggestion
		{
			name:      "single address with wrong port",
			addresses: []string{"localhost:63749"},
			wantErr:   true,
		},
```

We can add one more test to simulate Redis not being up or ready (using an incorrect port). This should cover your edge case

artempanko:
> Hi @artempanko , could you elaborate on what it is that is required here?

Just add restart on failure, like for PostgreSQL and typesense

PR #229, opened 2026-01-10
Fix typo in bug report template

myeljoud:


jerry-enebeli:
Thanks for the typo fix @myeljoud

PR #232, opened 2026-01-14
feat: further support `allow_overdraft` on GET requests on transactions

### LINKED ISSUE #221: allow_overdraft set to true is returned as false by GET and Search endpoints
**Describe the bug**
When `allow_overdraft` is set to true in the request payload, the GET and Search endpoints return allow_overdraft` as false instead.

**Expected behavior**
GET and Search endpoints should return the persisted value of `allow_overdraft`.

**Actual behavior**
GET and Search endpoints always return `allow_overdraft` = false, even when it was set to true.

shottah:
## What's Changed

Closes #221 

- [x] Added `allow_overdraft` field to the `transactions` table so we can persist historical requests to allow overdrafting on specific transactions when they are executed

> [!CAUTION] 
> - [x] 🚨 Added a migration script that will backfill all transactions to a default state of `allow_overdraft` to `FALSE`
>
> Recommending not to merge until a decision is made on backfilling historical data for `allow_overdraft` field

### DIFF
--- a/database/transactions_test.go
+++ b/database/transactions_test.go
@@ -69,7 +69,7 @@ func TestRecordTransaction_Success(t *testing.T) {
 	assert.NoError(t, err)
 
 	mock.ExpectExec("INSERT INTO blnk.transactions").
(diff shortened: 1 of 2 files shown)


shottah:
Based on other discussion this solution is pivoting away from supporting `allow_overdraft` on GET requests, to entirely removing it.

PR #279, opened 2026-04-23
Add inflight_commit_date for time-based auto-commit of inflight transactions

### LINKED ISSUE #277: Support for Guaranteed Future Settlements (Time-based Auto-commit)
I am looking for a way to perform guaranteed future settlements in Blnk. Currently, the system provides `scheduled_for` and `inflight` functionalities, but neither fully covers the use case of locking funds now for an automatic future settlement.

**Current Behavior & Limitations:**
* **`scheduled_for`:** Allows scheduling a future transaction, but **does not guarantee/lock funds** at the time of creation. The balance might be consumed by other operations between the scheduling and the execution time, leading to failures.
* **`inflight`:** Guarantees fund reservation immediately, but requires an external command (`commit`) to be finalized. It does not allow for **automatic time-based settlement**.

**The Need / Use Case:**
We need an operation that combines the best of both worlds: it should reserve/lock the balance immediately (like `inflight`) but automatically settle the transaction at a specific future date without requiring an external commit command. 

**Questions / Proposed Solution:**
1. Is there currently a parameter or a recommended approach in Blnk to achieve this guaranteed future settlement? 
2. If not, I would like to propose this as a feature. A possible implementation could be allowing `scheduled_for` to work alongside `inflight: true` (which would auto-commit on the scheduled date), or introducing a new parameter like `auto_commit_at` for inflight transactions.

rafaelcarvalhocaetano:
Closes #277

## What

Introduces `inflight_commit_date` as a new parameter for inflight transactions, enabling guaranteed future settlements without requiring an external commit command.

## Why

The existing `scheduled_for` does not lock funds at creation time, and `inflight` requires a manual commit. This feature combines both: reserves the balance immediately and automatically commits at a specified future date.

## Changes

- `model/transaction.go`: added `InflightCommitDate` field to the `Transaction` struct
- `api/model/transaction.go`: added `inflight_commit_date` to the `RecordTransaction` API model
- `api/model/model.go`: added date format validation for `inflight_commit_date` and parsing in `ToTransaction`
- `queue.go`: added `queueInflightCommit` and `QueueInflightCommit`, following the same pattern as `queueInflightExpiry`
- `transaction.go`: wires `QueueInflightCommit` in `QueueTransaction` when `inflight_commit_date` is set
- `config/config.go`: added `InflightCommitQueue` to `QueueConfig` and `LockWaitTimeout` default handling

## Tests

- `api/model/model_test.go`: validation and `ToTransaction` conversion for `inflight_commit_date`
- `config/config_test.go`: default and already-duration cases for `LockWaitTimeout`
- `queue_test.go`: skips enqueue when date is zero; enqueues task when date is set

### DIFF
--- a/api/model/model.go
+++ b/api/model/model.go
@@ -175,6 +175,14 @@ func (t *RecordTransaction) ValidateRecordTransaction() error {
 			return validateDateFormat("2006-01-02T15:04:05Z07:00", dateStr)
 		})),
 		),
+		validation.Field(&t.InflightCommitDate, validation.When(t.InflightCommitDate != "", validation.By(func(value interface{}) error {
+			dateStr, ok := value.(string)
+			if !ok {
+				return errors.New("invalid type for scheduled date")
+			}
+			return validateDateFormat("2006-01-02T15:04:05Z07:00", dateStr)
+		})),
+		),
 	)
 }
 
@@ -226,6 +234,7 @@ func (a *CreateAccount) ToAccount() model.Account {
 func (t *RecordTransaction) ToTransaction() *model.Transaction {
 	var scheduledFor time.Time
 	var inflightExpiryDate time.Time
+	var inflightCommitDate time.Time
 
 	if t.ScheduledFor != "" {
 		scheduledTime, err := time.Parse("2006-01-02T15:04:05Z07:00", t.ScheduledFor)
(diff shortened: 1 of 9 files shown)


jerry-enebeli:
Thank you for the PR @rafaelcarvalhocaetano, reviewing

PR #285, opened 2026-05-14
feat(identity): honor caller-supplied identity_id for deterministic creates

### LINKED ISSUE #286: CreateIdentity silently produces duplicate rows: server overwrites caller-supplied identity_id and there is no other dedup mechanism
## Summary

`POST /identities` cannot be made idempotent. Concurrent requests for the same external entity produce **two distinct identity rows** in `blnk.identity`, with no error surfaced to either caller. Blnk has no mechanism for deduplicating identities.

## Root cause

[`database/identity.go:46`](https://github.com/blnkfinance/blnk/blob/main/database/identity.go#L46) unconditionally overwrites whatever the caller sent:

\`\`\`go
identity.IdentityID = model.GenerateUUIDWithSuffix(\"idt\")
\`\`\`

Combined with:

- **No \`UNIQUE\` constraint on \`email_address\`, \`phone_number\`, or any name field** ([`sql/1708676327.sql:30-51`](https://github.com/blnkfinance/blnk/blob/main/sql/1708676327.sql#L30)) — only \`identity_id\` is \`UNIQUE\`, and the server generates that itself.
- **No SELECT-before-INSERT or \`ON CONFLICT\`** in `CreateIdentity`.
- **No idempotency-key header** support for the endpoint.

…there is no path by which a caller can express \"create this identity, or return the existing one if you've seen it before.\"

## Reproduction

\`\`\`bash
# Send the same payload twice in parallel
for i in 1 2; do
  curl -sX POST http://localhost:5001/identities \\
    -H \"X-blnk-key: \$KEY\" -H \"Content-Type: application/json\" \\
    -d '{
      \"identity_type\": \"individual\",
      \"first_name\": \"Alice\",
      \"last_name\":  \"Smith\",
      \"email_address\": \"alice@example.com\",
      \"category\": \"customer\",
      \"street\": \"x\", \"country\": \"x\", \"state\": \"x\", \"post_code\": \"x\", \"city\": \"x\"
    }' &
done; wait

# Both succeed with 201, returning two different idt_<uuid> values.
curl -s \"http://localhost:5001/search/identities\" -H \"X-blnk-key: \$KEY\" \\
  -d '{\"q\": \"alice@example.com\", \"query_by\": \"email_address\"}'
# Two hits.
\`\`\`

## Why this is load-bearing

Consumers building atop Blnk routinely need a 1:1 mapping between an entity in their system (a user, organization, holder of a wallet) and a Blnk identity. Today the only safe way to dedup is a **local mapping table** \`(external_id) → identity_id\` with a \`UNIQUE\` constraint, fronting every call to \`POST /identities\`. That defeats the value proposition of having identities live in the ledger — every consumer re-implements the same dedup machinery.

Worse, the failure is silent. A consumer that **thinks** they have dedup (e.g. because they pass a deterministic id and assume the server honors it) accumulates duplicates without any error signal until a downstream invariant catches it (mismatched balance counts, conflicting tokenized PII, etc.).

The official Go SDK ([`blnkfinance/blnk-go/identity.go`](https://github.com/blnkfinance/blnk-go/blob/main/identity.go)) does not expose \`identity_id\` on the request struct at all, so callers have no way to even attempt this pattern through the supported client.

## Docs vs. reality

The marketing copy in [\`identities/introduction.mdx\`](https://github.com/blnkfinance/blnk-docs/blob/main/identities/introduction.mdx#L186) advertises:

> By recording identities, it's easier to detect and prevent inconsistencies, **such as duplicate accounts** or incomplete credentials.

The create endpoint itself has no mechanism for either.

## Proposed fix

Two options, in order of preference:

1. **Honor caller-supplied \`identity_id\`** when present, leaving today's auto-generation path as the default. Maps PostgreSQL \`unique_violation\` to 409 Conflict so the caller can detect race-loss and fetch the existing row. Backward compatible. **PR opened: #285.**

2. **Idempotency-Key header support** on \`POST /identities\` — server-side store of \`(idempotency_key) → identity_id\` for replay. Heavier change but matches the pattern used elsewhere in the fintech ecosystem.

(1) is strictly less invasive and leverages the \`UNIQUE\` constraint that already exists on \`identity_id\`. (2) would be a larger architectural addition.

## Related

- PR: #285 (proposed fix for option 1)
- Affected SDK: \`blnkfinance/blnk-go\` — \`Identity\` struct has no \`IdentityID\` field, so even after the server change the SDK would need an update to expose it.

Shivam8584:
## Summary

`CreateIdentity` unconditionally overwrites `identity.IdentityID` with a freshly generated UUID ([`database/identity.go:46`](https://github.com/blnkfinance/blnk/blob/main/database/identity.go#L46)), so callers cannot derive a deterministic identity id from external state (e.g. a UUIDv5 of an external holder key) and rely on the existing `UNIQUE` constraint on `identity_id` to safely deduplicate concurrent creates.

Today two parallel requests for the same external entity produce two distinct identity rows in `blnk.identity` — the server has no other dedup mechanism (no `UNIQUE` on email/phone/name, no SELECT-before-INSERT, no `ON CONFLICT`), so duplicates accumulate silently. See #286 for the full bug report and reproduction.

This PR:

- **Preserves `identity.IdentityID` when supplied; generates one only when empty.** Fully backward compatible — existing callers send no id and get the same behaviour as today.
- **Validates that a supplied id carries the canonical `idt_` prefix** and rejects mismatches with `400 BAD_REQUEST`, keeping the id-format invariant the rest of the codebase assumes.
- **Maps PostgreSQL `unique_violation` on `identity_id` to `409 CONFLICT`** so callers using deterministic ids can detect "another concurrent request won the race" and fetch the existing row. This mirrors the established pattern in [`database/balance.go:274-278`](https://github.com/blnkfinance/blnk/blob/main/database/balance.go#L274) and [`database/ledger.go:62-63`](https://github.com/blnkfinance/blnk/blob/main/database/ledger.go#L62).

Closes #286.

## Test plan

- [x] `TestCreateIdentity_Success` — existing behaviour unchanged (empty id → generated)
- [x] `TestCreateIdentity_Fail` — existing error path unchanged
- [x] `TestCreateIdentity_CallerSuppliedID_Preserved` — supplied id round-trips
- [x] `TestCreateIdentity_CallerSuppliedID_BadPrefix` — non-`idt_` prefix → 400
- [x] `TestCreateIdentity_CallerSuppliedID_DuplicateConflict` — `unique_violation` → 409
- [x] `go build ./...` clean
- [x] Full `go test ./database/` clean (excluding pre-existing real-Postgres connection tests)

## Notes

No migration required — the `UNIQUE` constraint on `identity_id` already exists ([`sql/1708676327.sql:32`](https://github.com/blnkfinance/blnk/blob/main/sql/1708676327.sql#L32)).

### DIFF
--- a/database/identity.go
+++ b/database/identity.go
@@ -24,13 +24,28 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
+	"github.com/lib/pq"
+
 	"github.com/blnkfinance/blnk/internal/apierror"
 	"github.com/blnkfinance/blnk/internal/filter"
 	"github.com/blnkfinance/blnk/model"
 )
 
 // CreateIdentity inserts a new identity record into the database.
-// It generates a unique IdentityID, sets the creation timestamp, and stores the identity metadata.
+//
+// IdentityID handling:
+//   - If the caller supplies identity.IdentityID, it is preserved as-is.
+//     The value must carry the canonical "idt_" prefix followed by a valid
+//     UUID; otherwise the request is rejected with a 400. This lets callers
+//     derive deterministic identity ids (e.g. UUIDv5 of an external holder
+//     key) and rely on the UNIQUE constraint on identity_id for safe
+//     concurrent creation: parallel requests for the same external holder
+//     produce identical ids, one wins the insert, the others receive 409
+//     Conflict and can fetch the existing row.
+//   - If identity.IdentityID is empty, a fresh id is generated (existing
+//     behaviour).
+//
 // Parameters:
 // - identity: The identity object to be inserted.
 // Returns:
@@ -42,8 +57,17 @@ func (d Datasource) CreateIdentity(identity model.Identity) (model.Identity, err
 		return identity, apierror.NewAPIError(apierror.ErrInternalServer, "Failed to marshal metadata", err)
 	}
 
-	// Generate a unique identity ID and set the creation timestamp
-	identity.IdentityID = model.GenerateUUIDWithSuffix("idt")
+	// Honor a caller-supplied id when present; otherwise generate one.
+	if identity.IdentityID == "" {
+		identity.IdentityID = model.GenerateUUIDWithSuffix("idt")
+	} else {
+		if !strings.HasPrefix(identity.IdentityID, "idt_") {
+			return identity, apierror.NewAPIError(apierror.ErrBadRequest, "identity_id must start with the 'idt_' prefix", nil)
+		}
+		if _, err := uuid.Parse(strings.TrimPrefix(identity.IdentityID, "idt_")); err != nil {
+			return identity, apierror.NewAPIError(apierror.ErrBadRequest, "identity_id suffix after 'idt_' must be a valid UUID", err)
+		}
+	}
 	identity.CreatedAt = time.Now()
 
 	// Insert the identity record into the database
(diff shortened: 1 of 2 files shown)


Shivam8584:
@jerry-enebeli can you have a look?

jerry-enebeli:
Hi @Shivam8584 i'll review soon, thanks for the PR

jerry-enebeli:
Hi @Shivam8584!

Could we add validation to ensure the trailing ID after idt_ is a UUID and add test coverage for it as well? I'll merge once that's added.

Shivam8584:
Hi @jerry-enebeli — addressed your review feedback:

- Added `uuid.Parse` validation on the suffix after `idt_`; rejects with 400 if it isn't a valid UUID
- Added `TestCreateIdentity_CallerSuppliedID_BadUUID` covering empty suffix, non-UUID strings, truncated UUID, trailing garbage, and non-hex chars

I also force-pushed to re-author the commit under a GitHub-recognized email so the CLA matches cleanly (it now shows signed).

CI is in `action_required` on the new HEAD — could you click "Approve and run workflows" when you have a moment? Thanks!

PR #288, opened 2026-05-16
feat(transaction): allow skip_queue on the refund endpoint

Shivam8584:
## Problem

`POST /refund-transaction/:id` always queues the refund for asynchronous processing. There is no way for an API caller to request a **synchronous** refund — even though `Create Transaction` exposes `skip_queue` for exactly this case (time-sensitive operations needing immediate confirmation).

The core `RefundTransaction` already accepts a `skipQueue bool`, but `RefundWorker` passes `originalTxn.SkipQueue`. As the [Transaction Lifecycle docs](https://docs.blnkfinance.com/transactions/transaction-lifecycle#skip-transaction-queue) state, `skip_queue` is a request-time routing flag that is **deliberately not persisted**. A transaction fetched from storage for a refund therefore always deserializes `SkipQueue=false`, so the queued path is the only one reachable from the API.

## Changes

All additive and backward-compatible:

- **`RefundWorkerWithOptions(skipQueue bool) transactionWorker`** — returns a worker that refunds each job with an *explicit* `skipQueue` decision instead of inheriting the (always-false) persisted value. `RefundWorker` is left untouched.
- **`RefundTransaction` API handler** — optionally binds `{"skip_queue": true}` from the request body. An absent or empty body preserves the existing bodiless `POST /refund-transaction/:id` behavior (queued refund).
- **`TestRefundWorkerWithOptions`** — regression guard: a synchronous refund must apply immediately (`APPLIED`) even when the queued job's `SkipQueue` is `false`.

## Backward compatibility

The request body is optional. Existing callers that send no body get the unchanged queued-refund behavior. Only a caller that explicitly sends `{"skip_queue": true}` opts into synchronous processing.

## Testing

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt` — clean
- New test compiles and skips correctly under `-short` (it is a DB+Redis integration test, consistent with the existing `TestRefundWorkerFullFlow`).

### DIFF
--- a/api/transaction_api_test.go
+++ b/api/transaction_api_test.go
@@ -834,3 +834,102 @@ func TestInflightTransaction_Commit_WithAmount_API(t *testing.T) {
 	assert.Equal(t, originalPreciseAmount.Neg(originalPreciseAmount).String(), dbAfterFullCommit.Balance.String(), "Destination balance incorrect after full commit")
 	assert.Equal(t, int64(0), dbAfterFullCommit.InflightBalance.Int64(), "Destination inflight balance should be 0 after full commit")
 }
+
+// TestRefundTransaction_API exercises the POST /refund-transaction/:id
+// handler — specifically the optional skip_queue body added so callers
+// can request a synchronous refund.
+//
+// Three cases:
+//   - no body  → backward-compatible, refund is queued (default).
+//   - {"skip_queue": true} → refund is processed synchronously (APPLIED).
+//   - malformed body → 400, not a panic or a silent default.
+func TestRefundTransaction_API(t *testing.T) {
+	router, b, err := setupRouter()
+	if err != nil {
+		t.Fatalf("Failed to setup router: %v", err)
+	}
+	ctx := context.Background()
+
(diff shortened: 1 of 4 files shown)


Shivam8584:
Pushed a follow-up commit addressing two self-review points:

1. **Empty-body detection** — the first commit used `c.Request.ContentLength != 0`, which is unreliable (chunked requests report `-1`, proxies can desync the header). Switched to the canonical idiom: call `ShouldBindJSON` and treat **only** `io.EOF` (what `encoding/json` returns on an empty stream) as "no options supplied", rejecting any other bind error. Correct regardless of `Content-Length` / transfer encoding.

2. **Handler test** — the original test covered the worker; added `TestRefundTransaction_API` covering the HTTP handler directly: bodiless refund stays backward-compatible, `{"skip_queue": true}` applies synchronously, malformed body → 400.

Note on unknown fields: Blnk does not enable `EnableDecoderDisallowUnknownFields`, so a typo'd key (`{"skipqueue":true}`) is silently ignored — consistent with every other Blnk endpoint, so left as-is.

jerry-enebeli:
Thanks @Shivam8584, merging

PR #298, opened 2026-05-26
Harden API key authorization

Shivam8584:
## Summary

This PR fixes two API key authorization issues:

- an API key could manage keys for a different owner by passing another `owner` value in the request
- an API key could create a stronger key than itself by requesting broader scopes

After this change, non-master API keys are limited to managing their own keys and can only grant scopes they already have.

## What was wrong before

The API key endpoints were trusting request data from the caller instead of the authenticated API key.

That meant a caller could:

- create a key for some other owner
- list another owner's keys
- revoke another owner's keys
- mint a more privileged key and escalate access

## What this PR changes

For non-master API keys:

- the effective owner now comes from the authenticated API key, not the `owner` sent by the caller
- cross-owner key operations are rejected with `403`
- requested scopes must be covered by the caller's existing scopes

Master-key behavior is unchanged.

## Why this fix

This is a direct root-cause fix for the API key authz bug. It preserves the current API shape but removes the trust in caller-controlled owner and scope values.

## Validation

- `go test ./api/middleware -run 'TestScopeCovers|TestCanGrantScopes'`
- `go test ./api -run 'TestResolveAPIKeyOwner|TestValidateGrantedScopes'`

### DIFF
--- a/api/apikeys.go
+++ b/api/apikeys.go
@@ -1,12 +1,90 @@
 package api
 
 import (
+	"errors"
 	"net/http"
 
+	authz "github.com/blnkfinance/blnk/api/middleware"
 	"github.com/blnkfinance/blnk/api/model"
 	"github.com/blnkfinance/blnk/database"
+	coremodel "github.com/blnkfinance/blnk/model"
 	"github.com/gin-gonic/gin"
 )
+
+var (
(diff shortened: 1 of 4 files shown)


jerry-enebeli:
Thanks for identifying and fixing this @Shivam8584. I’ll merge it now. Expect a new release and a migration doc soon🙏

PR #299, opened 2026-05-26
Restrict hook management to the master key

Shivam8584:
## Summary

This PR locks hook management down to the master key.

After this change, ordinary API keys can no longer create, list, read, update, or delete hooks.

## What was wrong before

Hooks are currently global. They are not scoped to an owner or tenant.

That meant an API key with hook permissions could register an attacker-controlled webhook and then receive transaction payloads triggered by other tenants. In practice, this creates:

- cross-tenant data exfiltration
- an SSRF primitive through attacker-controlled hook URLs

## What this PR changes

- all hook management endpoints now require the master key
- non-master callers get `403` before any hook operation runs

## Why this fix

This is an intentional containment fix.

The deeper problem is that hooks need proper owner or tenant scoping across storage, lookup, and execution. That is a broader follow-up change. Until that exists, allowing API keys to manage global hooks is unsafe.

So this PR takes the safest short-term step: remove delegated hook management entirely and close the verified exploit path now.

## Follow-up

A future PR can safely restore delegated hook access only after hooks are scoped end to end.

## Validation

- `go test ./api -run 'TestEnsureHookManagementAuthorized'`

### DIFF
--- a/api/hooks.go
+++ b/api/hooks.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"net/http"
 
 	"github.com/blnkfinance/blnk/internal/apierror"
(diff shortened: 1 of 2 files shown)


jerry-enebeli (changes requested):
Hi @Shivam8584, thanks for the fix. Before merging, could you sign your commits with a verified signature?

PR #302, opened 2026-05-28
fix(workers): wire inflight commit queue consumer (fixes #300)

### LINKED ISSUE #300: Inflight commit queue jobs remain stuck in scheduled state
Re: PR #287 by @Shivam8584,

After configuring queue.inflight_commit_queue / BLNK_QUEUE_INFLIGHT_COMMIT, inflight commit jobs are successfully scheduled into Redis, but they are never picked up for processing.

**Steps to reproduce**
- Set queue.inflight_commit_queue in blnk.json or BLNK_QUEUE_INFLIGHT_COMMIT in .env
- Trigger a workflow that schedules an inflight commit
- Observe the queue state in Redis / monitoring dashboard

**Expected behaviour**
Scheduled inflight commit jobs should be consumed and processed by a worker.

**Actual behaviour**
Jobs remain indefinitely in the scheduled state and are never processed.

**Additional context**
- Verified from the monitoring dashboard that the jobs are successfully added to Redis
- The issue appears to be that no worker/process is consuming the inflight commit queue

Shivam8584:
## Summary

Commit `9bc2d24` added the full producer path for `inflight_commit_date` (model field, `queue.go` enqueue functions, `transaction.go` call site) but left the consumer side completely unwired in `cmd/workers.go`, causing scheduled commit jobs to sit in Redis in the `scheduled` state indefinitely.

### `config/config.go`
- Add `InflightCommitQueue: "new:inflight-commit"` to `defaultQueue`
- Add empty-string fallback in `setQueueDefaults()`

Without this, when `BLNK_QUEUE_INFLIGHT_COMMIT` is unset the field stays `""` and the producer silently enqueues to a nameless queue no server watches.

### `cmd/workers.go`
- Add `queues[cfg.Queue.InflightCommitQueue] = 1` in `initializeQueues()`
- Add `processInflightCommit` handler + `mux.HandleFunc` registration in `initializeTaskHandlers()`

Without the queue map entry the asynq server never polls the queue. Without the handler registration asynq archives every task with `"handler not found"`.

`processInflightCommit` calls `CommitInflightTransaction(ctx, txnID, big.NewInt(0))` — the zero-amount convention already used by `updateTransactionAmount` to commit the full remaining inflight balance.

## Test plan

```
go test ./cmd/ -run TestInflightCommit -v
```

- `TestInflightCommitQueueDefault` — field has a non-empty default after `setQueueDefaults()`
- `TestInflightCommitQueuePolled` — `initializeQueues()` includes the commit queue so the server polls it
- `TestInflightCommitHandlerRegistered` — mux has a handler registered (not "handler not found")

All three fail on `main` before this patch and pass after.

## Related

Closes #300
Producer added in: `9bc2d24` — `feat(transaction): add inflight_commit_date for time-based auto-commit`

### DIFF
--- a/cmd/workers.go
+++ b/cmd/workers.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"math/big"
 	"fmt"
 	"net/http"
 	"os/signal"
@@ -257,6 +258,22 @@ func (b *blnkInstance) indexBatchData(ctx context.Context, t *asynq.Task) error 
 	return nil
 }
 
+func (b *blnkInstance) processInflightCommit(ctx context.Context, t *asynq.Task) error {
+	var txnID string
+	if err := json.Unmarshal(t.Payload(), &txnID); err != nil {
+		logrus.Error(err)
+		return err
+	}
+
+	_, err := b.blnk.CommitInflightTransaction(ctx, txnID, big.NewInt(0))
+	if err != nil {
+		return err
+	}
+
+	logrus.Printf(" [*] Inflight Transaction Committed %s", txnID)
+	return nil
+}
+
 // processInflightExpiry handles the expiry of inflight transactions.
 // It voids the transaction by its ID and logs the action.
 func (b *blnkInstance) processInflightExpiry(cxt context.Context, t *asynq.Task) error {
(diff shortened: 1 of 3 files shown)


ShivamGupta-SM:
@jerry-enebeli check

jerry-enebeli:
Looks good, @Shivam8584. Merging now. Thanks.

PR #327, opened 2026-07-01
fix(reindex): include parent_transaction, precise_amount, precision in GetAllTransactions

### LINKED ISSUE #326: Reindex drops parent_transaction and precise_amount — GetAllTransactions SELECT omits them
### Summary

The Postgres→Typesense reindex path (`ReindexService.indexTransactions`) rebuilds the `transactions` collection via `Datasource.GetAllTransactions`, whose SQL `SELECT` omits `parent_transaction`, `precise_amount` (and `precision`). After any reindex these fields are therefore stored **empty** in Typesense, even though they are present in Postgres and set correctly by the live per-transaction indexing path.

This silently breaks:
- **Transaction lineage in search** — a committed inflight's resolving record has `parent_transaction` in the DB but not in the search index, so consumers can't link an inflight to its commit/void (or walk parent/child chains) via search.
- **`precise_amount` filter/facet** in search — the field is declared `string` in the schema but stored empty, so `filter_by=precise_amount:=...` matches nothing.

### Affected versions

Confirmed on **v0.14.5** and **v0.15.0** (identical query after the transaction-file refactor).

### Root cause

`GetAllTransactions` (v0.14.5 `database/transaction.go`; v0.15.0 `database/transaction_queries.go`):

```sql
SELECT transaction_id, source, reference, amount, currency, destination,
       description, status, hash, created_at, meta_data
FROM blnk.transactions ORDER BY created_at DESC LIMIT $1 OFFSET $2
```

No `parent_transaction`, `precise_amount`, or `precision`. The row scan populates only those columns, so `model.Transaction.ParentTransaction` / `.PreciseAmount` stay at their zero values; `toMap()` then emits them empty and `HandleNotification` indexes empty strings.

Meanwhile `getTransactionSchema()` **does** declare both fields (`{Name: "parent_transaction", ...}`, `{Name: "precise_amount", ...}`), and the live indexing path sets them — so live-indexed docs are correct and only **reindexed** docs are affected.

### Impact is amplified on ephemeral Typesense

When Typesense storage is ephemeral (e.g. Cloud Run with an in-memory `emptyDir` volume), the index is wiped and rebuilt from Postgres on every restart/scale event — so in such deployments these fields end up **permanently** empty, not just transiently after a manual reindex.

### Reproduction

1. Create an inflight transaction and commit it (Blnk emits a resolving record whose `parent_transaction` = the inflight's `transaction_id`).
2. Trigger a reindex (`POST /reindex`), or restart against an empty Typesense.
3. Search the transaction — `parent_transaction` and `precise_amount` come back empty, though both are present in Postgres.

### Suggested fix

Add `parent_transaction`, `precise_amount`, and `precision` to the `GetAllTransactions` `SELECT` and row scan, and ensure `toMap` carries them (converting `precise_amount` `big.Int` → string as the live path's `convertLargeNumbers` does), so reindexed docs match live-indexed docs.

### Related

Same class of "column omitted from a transaction query" bug as #292 (`POST /transactions/filter` omits `parent_transaction`).

sinkovsky:
## What

`GetAllTransactions` (`database/transaction_queries.go`) — the query the Postgres→Typesense reindex uses (`ReindexService.indexTransactions`) — omitted `parent_transaction`, `precise_amount`, and `precision` from its `SELECT` and row scan. This adds them, mirroring the existing single-row scan pattern in the package (`precise_amount` parsed via `parseBigInt`).

## Why

Because those columns weren't selected/scanned, every **reindexed** transaction document is written to Typesense with empty `parent_transaction` and `precise_amount`, even though Postgres has them and the live per-transaction indexing path sets them.

Impact:
- **Transaction lineage in search breaks** — a committed inflight's resolving record has `parent_transaction` in the DB but not in the index, so an inflight can't be linked to its commit/void via search.
- **`precise_amount` filter/facet returns nothing** — the field is declared `string` in the transaction schema but stored empty.

It's especially severe where Typesense storage is ephemeral (e.g. Cloud Run with an in-memory volume): the index is rebuilt from Postgres on every restart, so the fields are *permanently* empty there — not just after a manual reindex.

Fixes #326.

## How tested

- `go build ./...` — passes.
- `gofmt` — clean on both changed files.
- `go test -short -run GetAllTransactions ./database/` — passes. Updated the `TestGetAllTransactions_*` sqlmock rows to the new columns and added assertions that `ParentTransaction`, `PreciseAmount`, and `Precision` are now populated.
- Reproduced end-to-end on 0.14.5: with a persistent Typesense the fields are populated; after `POST /search/reindex` they drop to empty, confirming the reindex query as the cause.

_(The only failures in a full `go test -short ./database/` run are `TestConnectDB_Success` / `TestGetDBConnection_Singleton`, which need a live Postgres and are unrelated to this change.)_

### DIFF
--- a/database/transaction_queries.go
+++ b/database/transaction_queries.go
@@ -270,7 +270,7 @@ func (d Datasource) GetAllTransactions(ctx context.Context, limit, offset int) (
 
 	// Execute the query to retrieve all transactions
 	rows, err := d.Conn.QueryContext(ctx, `
-		SELECT transaction_id, source, reference, amount, currency, destination, description, status, hash, created_at, meta_data
+		SELECT transaction_id, source, reference, amount, precise_amount, precision, currency, destination, description, status, hash, created_at, meta_data, parent_transaction
 		FROM blnk.transactions
 		ORDER BY created_at DESC
 		LIMIT $1 OFFSET $2
@@ -292,6 +292,7 @@ func (d Datasource) GetAllTransactions(ctx context.Context, limit, offset int) (
 	for rows.Next() {
 		transaction := model.Transaction{}
 		var metaDataJSON []byte
+		var preciseAmountStr string
 
 		// Scan each row into the Transaction struct
 		err = rows.Scan(
@@ -299,6 +300,8 @@ func (d Datasource) GetAllTransactions(ctx context.Context, limit, offset int) (
 			&transaction.Source,
 			&transaction.Reference,
 			&transaction.Amount,
+			&preciseAmountStr,
+			&transaction.Precision,
 			&transaction.Currency,
 			&transaction.Destination,
 			&transaction.Description,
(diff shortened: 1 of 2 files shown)


jerry-enebeli:
Thanks for flagging this @sinkovsky, Reviewed and merged.

sinkovsky:
Thanks for prompt review @jerry-enebeli

PR #337, opened 2026-07-19
fix(reindex): include effective_date in GetAllTransactions

### LINKED ISSUE #335: effective_date lost on Postgres→Typesense reindex (0 in search) and null via REST — correct at write-time
## Summary

`effective_date` is honored and indexed correctly when a transaction is **created**, but it is lost on the read paths:

1. The **Postgres→Typesense reindex** (on boot) resets `effective_date` to `0` in the search index — the real value is dropped from search after any restart.
2. **`GET /transactions/:id`** returns `"effective_date": null` even when the DB row has a value.

Postgres retains the correct value throughout; only the read paths lose it.

This is the same class of reindex field-loss fixed in #327 (which added `parent_transaction` + `precise_amount` to `GetAllTransactions`) — `effective_date` appears to be missing from that same reindex SELECT + Typesense document mapping, plus the single-transaction REST serializer.

## Reproduce (verified on v0.15.1)

1. Create a transaction with a backdated `effective_date`:
```
POST /transactions
{ "precise_amount": 500, "precision": 100, "reference": "eff-1", "currency": "USD",
  "source": "@world", "destination": "<balance>", "description": "x",
  "allow_overdraft": true, "skip_queue": true,
  "effective_date": "2020-06-15T12:00:00Z" }
```

2. Search immediately — **correct**:
```
POST /search/transactions { "q": "eff-1", "query_by": "reference" }
→ document.effective_date = 1592222400   ✓ (2020-06-15)
```

3. Restart the server (triggers the Postgres→Typesense reindex), search again — **wrong**:
```
→ document.effective_date = 0            ✗
```
Meanwhile Postgres still holds `2020-06-15 12:00:00`, and `GET /transactions/<id>` returns `"effective_date": null`.

## Impact

Deployments with **ephemeral Typesense** (rebuilt from Postgres on every restart) permanently lose `effective_date` from search — value-dated / backdated transactions can't be sorted or displayed by their business date, and REST can't recover it either. `created_at` (system insert time) is the only date available through the API.

## Fix

Add `effective_date` to:
- the reindex `GetAllTransactions` SELECT + the Typesense document mapping (same location as #327), and
- the `GET /transactions/:id` response serializer.

Happy to open a PR — it mirrors #327.

Version: v0.15.1

sinkovsky:
Fixes #335.

## Problem

`GetAllTransactions` (`database/transaction_queries.go`) is the Postgres source for the Typesense reindex, but its SELECT omitted `effective_date`. So on every reindex — which deployments with ephemeral Typesense run on each restart — transactions lose their business/value date: search returns `effective_date=0` even for records that had one set at write time.

This is the same class of reindex field-loss fixed in #327 (which added `parent_transaction` + `precise_amount` to this exact query); `effective_date` was simply left out.

## Fix

Add `effective_date` to the SELECT and scan it via `sql.NullTime` (nil for old records, where `GetEffectiveDate()` already falls back to `created_at`). One hunk in `GetAllTransactions` — the search-doc builder (`toMap`) and the Typesense schema already handle the field; the reindex just wasn't populating it.

## Verify

1. Create a transaction with a backdated `effective_date`.
2. Search immediately → `document.effective_date` is correct.
3. Restart the server (triggers the Postgres→Typesense reindex).
4. Before this fix: search shows `effective_date = 0`. After: the reindexed doc retains the real value.

`go build ./database/` passes.

### DIFF
--- a/database/transaction_queries.go
+++ b/database/transaction_queries.go
@@ -270,7 +270,7 @@ func (d Datasource) GetAllTransactions(ctx context.Context, limit, offset int) (
 
 	// Execute the query to retrieve all transactions
 	rows, err := d.Conn.QueryContext(ctx, `
-		SELECT transaction_id, source, reference, amount, precise_amount, precision, currency, destination, description, status, hash, created_at, meta_data, parent_transaction
+		SELECT transaction_id, source, reference, amount, precise_amount, precision, currency, destination, description, status, hash, created_at, effective_date, meta_data, parent_transaction
 		FROM blnk.transactions
 		ORDER BY created_at DESC
 		LIMIT $1 OFFSET $2
@@ -293,6 +293,7 @@ func (d Datasource) GetAllTransactions(ctx context.Context, limit, offset int) (
 		transaction := model.Transaction{}
 		var metaDataJSON []byte
 		var preciseAmountStr string
+		var effectiveDate sql.NullTime
 
 		// Scan each row into the Transaction struct
 		err = rows.Scan(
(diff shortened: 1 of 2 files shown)


jerry-enebeli:
Thanks @sinkovsky merging

PR #346, opened 2026-08-02
fix: clear terminal inflight metadata

### LINKED ISSUE #293: A committed/voided transaction permanently reports inflight: true
# Bug: a committed/voided transaction permanently reports `inflight: true`

**Type:** bug
**Component:** `inflight_transaction.go` — `finalizeCommitment`, `finalizeVoidTransaction`
**Version checked:** `main` @ `e0f8676` (2026-05-18)

## Summary

When an inflight transaction is committed (or voided), the resulting
terminal transaction has `status: APPLIED` (or `VOID`) but still
reports `inflight: true`. The wrong value is returned by the
commit/void call **and** by every subsequent `GET /transactions/{id}`
and `GET /transactions/reference/{ref}` — it is not transient.

A transaction cannot be both `APPLIED` and `inflight`. Any consumer
that classifies a transaction by the `inflight` flag (rather than
`status`) will misjudge a settled transaction as still in flight.

## Root cause

`inflight` is **not a column** on `blnk.transactions` — it is stored
inside the `meta_data` JSONB. An inflight transaction's `meta_data`
therefore contains `{"inflight": true, ...}`.

`finalizeCommitment` (`inflight_transaction.go:378`) builds the
committed child by mutating the inflight transaction in place:

```go
func (l *Blnk) finalizeCommitment(ctx context.Context, transaction *model.Transaction, withQueue bool) (*model.Transaction, error) {
    transaction.Status = StatusCommit
    transaction.ParentTransaction = transaction.TransactionID
    transaction.CreatedAt = time.Now()
    ...
    transaction.TransactionID = model.GenerateUUIDWithSuffix("txn")
    transaction.Reference = model.GenerateUUIDWithSuffix("ref")
    transaction.Hash = transaction.HashTxn()
    ...
}
```

It overrides status / id / reference / hash / parent — but **never
touches `transaction.MetaData`**, which still holds `inflight: true`.
The committed child is then INSERTed with that metadata verbatim.
`finalizeVoidTransaction` (`inflight_transaction.go:566`) has the same
omission.

On read, `restoreTransactionFlagsFromMetadata` (`transaction.go:1653`)
re-derives the struct field from metadata:

```go
func restoreTransactionFlagsFromMetadata(transaction *model.Transaction) {
    if v, ok := transaction.MetaData["inflight"].(bool); ok {
        transaction.Inflight = v          // <-- a committed txn becomes Inflight=true
    }
    ...
}
```

So a terminal transaction's `meta_data` carries `inflight: true`, and
every read faithfully restores `Inflight = true` from it.
`transformTransaction` strips the `inflight` key out of the response
`meta_data` but leaves `result.Inflight = true`.

## Steps to reproduce

```bash
# 1. Create an inflight transaction.
INF=$(curl -s -X POST http://localhost:5001/transactions \
  -H 'X-Blnk-Key: <key>' -H 'Content-Type: application/json' \
  -d '{"amount":100,"precision":100,"currency":"USD","reference":"repro-'"$(date +%s)"'",
       "source":"<balance_id>","destination":"@World","description":"x",
       "inflight":true,"skip_queue":true,"allow_overdraft":true}')
INFID=$(echo "$INF" | jq -r .transaction_id)

# 2. Commit it — inspect the response.
COMMIT=$(curl -s -X PUT "http://localhost:5001/transactions/inflight/$INFID" \
  -H 'X-Blnk-Key: <key>' -H 'Content-Type: application/json' \
  -d '{"status":"commit"}')
echo "$COMMIT" | jq '{status, inflight}'
#   -> { "status": "APPLIED", "inflight": true }      ❌ BUG

# 3. GET it back later — still wrong.
CHILDID=$(echo "$COMMIT" | jq -r .transaction_id)
curl -s "http://localhost:5001/transactions/$CHILDID" -H 'X-Blnk-Key: <key>' | jq '{status, inflight}'
#   -> { "status": "APPLIED", "inflight": true }      ❌ BUG
```

### Confirmed against the database

The committed child's stored `meta_data` (queried directly from
Postgres) carries the inherited flag — there is no `inflight` column,
and the value survives a full Redis `FLUSHALL`, so it is genuinely
persisted, not cached:

```
txn_<parent> | INFLIGHT | {"inflight": true, "allow_overdraft": true}
txn_<child>  | APPLIED  | {"inflight": true, "allow_overdraft": true}   <-- stale
```

### Expected

A committed transaction should report `inflight: false` (status
`APPLIED`); a voided one `inflight: false` (status `VOID`). The
`inflight` flag must reflect the terminal state, not be inherited from
the parent.

## Suggested fix

In `finalizeCommitment` and `finalizeVoidTransaction`, after the
field overrides, clear the inherited flag from metadata so the
persisted child and every subsequent read are correct:

```go
if transaction.MetaData != nil {
    delete(transaction.MetaData, "inflight")
}
```

Clearing it in `meta_data` is the right layer because
`restoreTransactionFlagsFromMetadata` derives the `Inflight` struct
field *from* metadata — the metadata is the source of truth, and it is
what is wrong. (Setting `transaction.Inflight = false` alone would be
overwritten on the next read.)

## Note

This is the same class of bug as #291 (`prepareRefundTransaction`):
a terminal/derived transaction is built by copying a source transaction
and overriding a partial set of fields, leaving one field stale. The
commit/void path and the refund path share this pattern.

Guflly:
Clear the inflight marker from committed and voided transactions after outbox preparation.

Waiting until the outbox is prepared keeps the existing inflight lineage behavior intact while preventing terminal transactions from retaining the marker.

Fixes #293

Tests:
- `go test -run "TestClearTerminalInflightMetadata|TestBuildTransactionExecutionWorkClearsTerminalInflightMetadata" .`
- `go vet .`
- `go build .`

### DIFF
--- a/transaction_execution.go
+++ b/transaction_execution.go
@@ -728,11 +728,14 @@ func (l *Blnk) buildTransactionExecutionWork(ctx context.Context, transaction *m
 		}, true
 	}
 
+	outbox := l.prepareTransactionOutbox(ctx, transaction, sourceBalance, destinationBalance)
+	clearTerminalInflightMetadata(transaction)
+
 	return queuedBatchPostCommitWork{
 		transaction:        transaction,
 		sourceBalance:      sourceBalance,
 		destinationBalance: destinationBalance,
-		outbox:             l.prepareTransactionOutbox(ctx, transaction, sourceBalance, destinationBalance),
+		outbox:             outbox,
 	}, false
 }
 
--- a/transaction_inflight.go
+++ b/transaction_inflight.go
@@ -83,6 +83,17 @@ func IsInflightTransaction(transaction *model.Transaction) bool {
 			transaction.MetaData["inflight"] == true)
 }
 
+func clearTerminalInflightMetadata(transaction *model.Transaction) {
+	if transaction == nil || transaction.Status == StatusInflight {
+		return
+	}
+
+	transaction.Inflight = false
+	if transaction.MetaData != nil {
+		delete(transaction.MetaData, "inflight")
+	}
+}
+
 // CommitWorker processes commit transactions from the jobs channel and sends the results to the results channel.
 // It starts a tracing span, processes each transaction, and records relevant events and errors.
 //
@@ -394,6 +405,7 @@ func (l *Blnk) finalizeCommitment(ctx context.Context, transaction *model.Transa
 	defer span.End()
 
 	transaction.Status = StatusCommit
+	transaction.Inflight = false
 	transaction.ParentTransaction = transaction.TransactionID
 	transaction.CreatedAt = time.Now()
 	if transaction.EffectiveDate == nil {
@@ -422,6 +434,7 @@ func (l *Blnk) finalizeCommitment(ctx context.Context, transaction *model.Transa
 		return transaction, nil
 	}
 
+	clearTerminalInflightMetadata(transaction)
 	transaction.Status = StatusApplied
 
 	span.AddEvent("Commitment finalized", trace.WithAttributes(attribute.String("transaction.id", transaction.TransactionID)))
@@ -593,6 +606,7 @@ func (l *Blnk) finalizeVoidTransaction(ctx context.Context, transaction *model.T
 	defer span.End()
 
 	transaction.Status = StatusVoid
+	transaction.Inflight = false
 	transaction.PreciseAmount = amountLeft
 	transaction.CreatedAt = time.Now()
 	if transaction.EffectiveDate == nil {
--- a/transaction_inflight_async_test.go
+++ b/transaction_inflight_async_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package blnk
 
 import (
+	"context"
 	"errors"
 	"math/big"
 	"testing"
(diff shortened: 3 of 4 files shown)


Shivam8584:
I filed #293, so I pulled this down and tested it against `main`. It fixes what I reported, and the approach is right — a few notes below, one of which is a gap worth a follow-up.

## Verified against a live server

Applied the patch to `main` (`c8c003c`), ran a real Postgres + Redis, exercised the inflight lifecycle end to end:

| case | result |
|---|---|
| inflight parent, still open | `INFLIGHT` / `inflight: true` ✅ |
| committed child (sync, `skip_queue`) | `APPLIED` / `inflight: false` ✅ |
| lineage_outbox rows | parent = 1, child = **0** ✅ |
| `go test ./api/ ./model/ .` | green, no regressions ✅ |

## Ordering the clear after the outbox is the important bit

Worth spelling out for reviewers, because the obvious fix is wrong in a way that's easy to miss.

The suggestion in my own issue was to `delete(transaction.MetaData, "inflight")` inside `finalizeCommitment`. I tried exactly that before finding this PR, and it **silently duplicates fund lineage**:

```
with delete() in finalizeCommitment:
  status: APPLIED | inflight: false          ← bug "fixed"
  lineage rows for the committed child: 1    ← was 0

blnk.lineage_outbox then holds BOTH:
  txn_0dbe…  INFLIGHT   ← created when the hold was placed
  txn_0119…  APPLIED    ← duplicate, same money tracked twice
```

`prepareTransactionOutbox` identifies inflight commits by that very key (`transaction_execution.go:766`), and status can't stand in for it because `updateTransactionDetails` has already mapped `COMMIT → APPLIED` by the time the outbox is prepared. Capturing `outbox` first and clearing after is what keeps that guard working — which is exactly what this PR does.

I confirmed the queued path too (`finalizeCommitment(withQueue: true)` enqueues at what is now line ~423 *before* the clear at ~437, so the serialized payload still carries the marker) — but the worker routes through `buildTransactionExecutionWork`, so it gets cleared before persistence. Covered.

## Gap: `RejectTransaction` isn't reached

`RejectTransaction` persists via `datasource.RecordTransaction` directly (`transaction_rejection.go:48`), bypassing `buildTransactionExecutionWork`, so a **rejected** inflight transaction keeps `meta_data.inflight = true` and goes on reporting `inflight: true` — the same bug on a different terminal status.

It's reachable: a queued inflight transaction that fails on insufficient funds or exhausts its retries is rejected at `cmd/workers.go:156` and `queue_recovery.go:179`.

`clearTerminalInflightMetadata` already handles this correctly (its guard is `Status == StatusInflight`, and a rejected row is `REJECTED`), so it looks like one line before the persist:

```go
transaction.MetaData["blnk_rejection_reason"] = reason
clearTerminalInflightMetadata(transaction)

transaction, err := l.datasource.RecordTransaction(ctx, transaction)
```

Happy either way on whether that belongs here or in a follow-up — it's a different code path and this PR is already correct on its own.

## One test suggestion

A regression test asserting the committed child has **zero** `lineage_outbox` rows (and the parent one) would pin the ordering. Without it, a future tidy-up that moves `clearTerminalInflightMetadata` back above `prepareTransactionOutbox` still passes every existing test while quietly double-counting lineage.

Nice fix — the ordering here is subtler than it looks.

jerry-enebeli:
Thank you, @Shivam8584 for the reviews; I agree with them. 

We should make the Gap a separate issue/PR.

jerry-enebeli:
Thank you for the fix @Guflly

PR #347, opened 2026-08-03
fix(hooks): return all hooks when type is omitted

### LINKED ISSUE #345: GET /hooks without a type query returns an empty list
**Describe the bug**

`GET /hooks` returns `200 OK` with an empty array when no `type` query parameter is provided, even when PRE_TRANSACTION and/or POST_TRANSACTION hooks exist.

Filtered requests such as `GET /hooks?type=PRE_TRANSACTION` work correctly.

**To Reproduce**

1. Register a PRE_TRANSACTION or POST_TRANSACTION hook using `POST /hooks`.
2. Confirm it exists using:
   ```
   GET /hooks?type=PRE_TRANSACTION
   ```
   or:
   ```
   GET /hooks?type=POST_TRANSACTION
   ```
3. Request all hooks without a query:
   ```
   GET /hooks
   ```
4. Observe that the response is `200 OK` with:
   ```json
   []
   ```

**Expected behavior**

`GET /hooks` without a `type` query should return hooks from both the PRE_TRANSACTION and POST_TRANSACTION sets.

Niketion:
`GET /hooks` passed an empty hook type to `getTypeKey`, causing the manager to read the unused `hooks:` set and return an empty result.

When no type is provided, this change reads both the pre- and post-transaction hook sets, removes duplicate IDs, and loads hooks through the existing `GetHook` path. Filtered listing behavior remains unchanged.

Fixes #345

Tests:
- `go test ./internal/hooks -count=1` — passed using a local Windows `-exec` launcher because the host blocks Go test binaries executed from the temporary directory
- `go test ./internal/hooks -run TestListHooks -count=50 -shuffle=on` — passed with the same launcher
- `go vet ./internal/hooks`
- `git diff --check`

### DIFF
--- a/internal/hooks/manager.go
+++ b/internal/hooks/manager.go
@@ -221,10 +221,28 @@ func (m *redisHookManager) GetHook(ctx context.Context, hookID string) (*Hook, e
 // - []*Hook: A slice of retrieved hooks.
 // - error: An error if retrieval fails.
 func (m *redisHookManager) ListHooks(ctx context.Context, hookType HookType) ([]*Hook, error) {
-	typeKey := getTypeKey(hookType)
-	hookIDs, err := m.client.SMembers(ctx, typeKey).Result()
-	if err != nil {
-		return nil, err
+	var hookIDs []string
+	if hookType == "" {
+		seen := make(map[string]struct{})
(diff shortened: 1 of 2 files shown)


jerry-enebeli:
Thanks for the fix @Niketion

Issue #330, opened 2026-07-13
Feature: dry_run (preview) mode for the transaction APIs — project resulting balances without persisting


Shivam8584:
# Feature: `dry_run` (preview) mode for the transaction APIs — project resulting balances without persisting

## Summary

Blnk already supports `dry_run` for **reconciliation** (`model/reconciliation_model.go` `IsDryRun`, guarded in `reconciliation.go`), but the **transaction** APIs have no equivalent. There is no first-class way to answer:

> *"If I post this transaction, would it succeed, and what would the resulting balances be?"*

…without actually creating an (immutable) transaction.

Because Blnk models money as **immutable flows** and corrections are handled through **new entries, never edits**, a mistaken post can only be undone with another entry. That makes a **read-only preview** valuable for pre-flight UI checks, quoting, back-office correction/adjustment previews, and for testing overdraft / inflight / distribution logic.

Today integrators must either (a) post and then reverse if wrong, or (b) re-derive Blnk's own balance math client-side (`available = balance − inflight_debit_balance`, overdraft rules, precision, multi-leg distribution) — which is racy and duplicates ledger logic that should live in the ledger.

This proposes a **`dry_run: true`** flag that runs the **full existing evaluation path** and returns the **projected** balances, changing nothing.

---

## Why this is a small, idiomatic change

The apply pipeline already computes new balances **in-memory before it persists**. In `recordTransactionSingle` (`transaction_execution.go`), the flow is:

```
resolveBalanceIDs → getSourceAndDestination (load balances)
  → processBalances → model.UpdateBalances   ← computes new balances + runs canProcessTransaction (overdraft/insufficiency)
  ────────────── read-only seam ──────────────
  → buildTransactionExecutionWork → persistSingleTransactionExecutionWork   ← first write
  → runTransactionPostCommitWork (webhooks, monitors, index)
```

Right after `processBalances` returns, the `source` / `destination` balance structs already hold the **projected post-transaction values**, and overdraft/insufficiency has already been decided. A dry-run simply **returns at that seam** with the projected balances (and any projected rejection), skipping persistence, queueing, and webhooks.

This mirrors the existing reconciliation `dry_run` idiom: thread one flag, guard every write/side-effect with it — rather than adding a parallel code path.

---

## API surface

`dry_run` is honored on every balance-affecting write. In all cases a dry-run responds **`200 OK`** (nothing was created), never `201`.

| Endpoint | Method | Dry-run behavior |
|---|---|---|
[...]

PR #348, opened 2026-08-09
feat(transactions): dry_run preview for create and refund

Shivam8584:
Adds `dry_run` to `POST /transactions` and `POST /refund-transaction/:id`, answering "if I post this, would it succeed and what would the balances become" without writing anything.

Part one of two, per #330 — bulk and inflight follow in a second PR, where the cross-leg accumulator and multi-leg parent expansion carry most of the complexity. Corrections to the original issue text are in [this comment](https://github.com/blnkfinance/blnk/issues/330#issuecomment-5229697928).

```jsonc
POST /transactions   { "dry_run": true, "amount": 120, ... }   → 200

{
  "dry_run": true,
  "would_apply": true,
  "precise_amount": "12000",
  "balances": [
    { "role": "source",      "current_balance": "50000", "resulting_balance": "38000", ... },
    { "role": "destination", "current_balance": "0",     "resulting_balance": "12000", ... }
  ]
}
```

### How it works

The projection runs **the same arithmetic a real post runs** — `processBalances`, and through it `UpdateBalances` and `canProcessTransaction` — against deep copies of the balances. Sharing the math rather than reimplementing it is the whole point: a preview that computed availability its own way would drift from enforcement the first time overdraft rules changed, and a preview that disagrees with reality is worse than none.

Three things the code made necessary that aren't obvious:

- **`POST /transactions` doesn't reach `RecordTransaction`.** It goes through `QueueTransaction`, which persists a `QUEUED` row and applies balances later in a worker. The projection drives the apply seam directly instead — threading a flag through the queue would still write a row.
- **Resolving an `@indicator` writes.** `getOrCreateBalanceByIndicator` creates the balance when it's missing, so resolution goes through read-only lookups and an unknown indicator is projected against a zeroed virtual balance rather than created.
- **The apply path mutates in place.** `UpdateBalances` writes `PreciseAmount` onto the transaction, and `big.Int`'s `Add`/`Sub` mutate their receiver — so a plain struct copy would share the underlying integers and a "before" snapshot would silently show "after" values. Hence `Clone()` in the first commit.

Locks are taken so both balances are read as of one moment, but acquired directly rather than through `executeWithLock`: a preview shouldn't record hot-pair contention and steer hot-lane routing for real traffic.

### Deliberate choices

**No new error codes.** A projected rejection is a `200` with `would_apply: false` — the request succeeded, the answer is "no". The rejection carries the same code a real post returns, resolved through the existing `classifyMessage`, so client-side handling for `TXN_INSUFFICIENT_FUNDS` works against a preview unchanged. `codes.go` is untouched.

**`dry_run` stays out of `model.Transaction`.** It describes what to do with the request, not the movement, so it lives at the transport boundary — which also makes it structurally impossible for a preview transaction to reach the writer by accident.

**Where the ledger is permissive, the preview reports rather than invents a rule.** The apply path performs no currency check, so a mismatched currency is projected with a note instead of rejected. A preview stricter than enforcement would be misleading.

**`dry_run` wins over `skip_queue`**, and always answers synchronously with `200`.

### Scope of the guarantee

A dry run changes no ledger state: no transaction row, no balance mutation, no queue entry, no webhook, no hook, and the reference is not consumed.

It is *not* byte-for-byte write-free at the HTTP layer — auth middleware still updates `api_keys.last_used_at`, and OTel spans are still emitted. Both are deliberate; a preview you can't trace is worse to operate.

One known wart: because permissions are gated by HTTP method rather than by effect, a preview on `POST /transactions` still requires write scope — awkward for the pre-flight UI check that motivates the feature. Happy to follow up separately if you'd like previews to be read-scope-accessible; it seemed like a permission-model change that didn't belong in a feature PR.

### Tests

26 new tests. The ones that matter:

- **Changes nothing** — balances byte-identical including `version` (the optimistic-lock counter moves on *any* persisted write, so it's the sharpest tripwire); no transaction row; reference still usable by a real post afterwards; unknown `@indicator` not created. The service-level tests run against `sqlmock`, so an unexpected `INSERT` fails the test by itself rather than needing an assertion.
- **`Clone` doesn't alias** — every `*big.Int` checked with `NotSame`, plus mutate-the-clone-assert-the-original.
- **Projection matches enforcement** — insufficient funds, inflight-adjusted availability, zero amount rejected (validate runs inside `UpdateBalances`, so it errors rather than being a silent no-op).
- **Bodiless refund still works** — the new field is optional and `POST /refund-transaction/:id` with an empty body keeps meaning "refund normally".

Verified end to end against a running server: previewed a transaction, then really posted it — the projection matched the resulting balances field for field, and `version` was unchanged across the preview.

### DIFF
--- a/api/model/transaction.go
+++ b/api/model/transaction.go
@@ -44,6 +44,16 @@ type RecordTransaction struct {
 	Destinations       []model.Distribution   `json:"destinations"`
 	MetaData           map[string]interface{} `json:"meta_data"`
 	EffectiveDate      *time.Time             `json:"effective_date,omitempty"`
+
+	// DryRun projects the transaction's effect on the source and destination
+	// balances and returns the result without applying it. No transaction row
+	// is written, no balance changes, nothing is queued, no webhook fires, and
+	// the reference is not consumed.
+	//
+	// A dry run always responds synchronously with 200 rather than 201, and
+	// takes precedence over skip_queue: a projection is inherently an immediate
+	// answer about what a real post would do.
+	DryRun bool `json:"dry_run"`
 }
 
 // BulkTransactionRequest is the public API request shape for creating a batch
--- a/api/transaction_dryrun_api_test.go
+++ b/api/transaction_dryrun_api_test.go
@@ -1,0 +1,280 @@
+package api
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	blnk "github.com/blnkfinance/blnk"
+	model2 "github.com/blnkfinance/blnk/api/model"
+	"github.com/blnkfinance/blnk/internal/request"
+	"github.com/blnkfinance/blnk/model"
+)
+
+// newDryRunFixture creates a ledger with a funded source balance and an empty
+// destination, and returns their ids.
+func newDryRunFixture(t *testing.T, b *blnk.Blnk) (string, string) {
+	t.Helper()
+
+	ledger, err := b.CreateLedger(model.Ledger{Name: gofakeit.Name()})
+	require.NoError(t, err)
+
+	source, err := b.CreateBalance(t.Context(), model.Balance{LedgerID: ledger.LedgerID, Currency: "USD"})
+	require.NoError(t, err)
+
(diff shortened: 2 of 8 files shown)


jerry-enebeli:
Hi @Shivam8584 taking a look

Gee-Hive (approved):
Hi @Shivam8584 , I have reviewed and this looks good. 

For future PRs it would be nice to set signed commits, because currently your commits don't have verified signatures. Thanks.

PR #349, opened 2026-08-09
feat(transactions): dry_run preview for bulk and inflight

Shivam8584:
Completes #330 by adding `dry_run` to `POST /transactions/bulk` and `PUT /transactions/inflight/:txID`.

**Stacked on #348** — please merge that first; this branch contains its three commits, so the diff here is just the final commit. Rebases to `main` cleanly once #348 lands.

### Bulk: the projection mode has to match how the batch really runs

This is the part the original issue got wrong, and it's the main reason this is a separate PR.

I'd claimed later items see earlier items' balance changes. That's only true with `skip_queue: true`. `processBulkTransactions` assigns the batch-level flag to every item (`transaction_bulk.go:36`), so on the default path each item goes to `processTransactionAsync`'s detached goroutine and the loop moves on — the items race.

So the mode follows the flag, and the response says which one you got:

```jsonc
// skip_queue: true — 500.00 available, items of 450.00 then 100.00
{ "cumulative": true, "would_apply": false,
  "results": [ { "would_apply": true },
               { "would_apply": false, "rejection": { "code": "TXN_INSUFFICIENT_FUNDS" } } ] }

// skip_queue: false — same input, both fit on their own
{ "cumulative": false, "would_apply": true,
  "results": [ { "would_apply": true }, { "would_apply": true } ] }
```

Identical input, opposite verdict, because the execution semantics genuinely differ. Projecting cumulatively on the default path would assert an ordering the ledger doesn't provide — an authoritative-looking wrong answer, which is worse than no preview.

Batch-level `balances` are reported only in cumulative mode: without a guaranteed order there's no single combined outcome to state.

**No `all_or_nothing` field**, for the same honesty reason. `rollbackBatchTransactions` compensates by voiding or refunding already-applied items *after* a failure, and that compensation can itself fail. The response carries a note saying so rather than implying isolation.

Balances are fetched once per distinct balance and reused across items — needed for correctness in cumulative mode anyway, and it keeps a 10k-item batch from doing 20k round trips.

### Inflight: reuses the existing expansion rather than reimplementing it

`preValidateInflightAction` already handles the case where an id names the parent of a split rather than a single hold, so the projection resolves legs the same way and projects **every** leg — a per-transaction preview on a parent would otherwise answer for one representative leg, or fail outright.

Void has no partial form: `finalizeVoidTransaction` always uses the full remaining hold. An amount sent with a void is reported as ignored rather than silently dropped:

```jsonc
PUT /transactions/inflight/:id  { "dry_run": true, "status": "void", "precise_amount": 4000 }

{ "operation": "void", "precise_amount": "10000",
  "notes": ["amount is ignored when voiding; a void always releases the full remaining hold"] }
```

Both projections run on clones — `validateAndUpdateAmount` writes the settled amount onto the transaction it's given. The dry-run branch also sits ahead of `ApplyPrecisionWithDBLookup`, which caches into package-level state.

### Tests

9 new tests covering: cumulative mode catching an intra-batch shortfall; independent mode not assuming order; batch and inflight previews writing nothing (balances and `version` unchanged, hold still `INFLIGHT`, no settlement rows); void ignoring its amount; and unsupported actions still being rejected as malformed.

Verified against a running server — the cumulative case correctly rejected the second item at `TXN_INSUFFICIENT_FUNDS` where independent mode passed both, the commit projection showed the hold releasing *and* the money moving while void showed the hold releasing with the balance unchanged, and afterwards the hold was still `INFLIGHT` with zero settlement rows written.

---

### Review follow-ups

Three points raised by @Gee-Hive, all fixed on this branch:

**1. A dry-run commit above the original amount reported `would_apply: true`.** The dry-run branch passed only `req.PreciseAmount` to `PreviewInflightAction`, skipping the `amount` → precision → `precise_amount` resolution the real commit path does a few lines below. With `precise_amount` unset that arrived as `nil`, which is the sentinel for "commit the full remaining amount", so an oversized `amount` was discarded rather than validated. The resolution now sits above the `dry_run` check, so both branches share it.

**2. Inflight dry-run omitted `status` and `amount`.** `PreviewInflightAction` built its own `TransactionPreview` literal instead of going through `newPreviewFor`, so both fields sat at their zero values. `status` is now the status the settlement would carry once applied (`APPLIED` for a commit, `VOID` for a void — not the internal `COMMIT`/`VOID` marker), and `amount` is the float rendering of the precise total already reported.

**3. A rejected projection still reported `status: "APPLIED"`.** `status` is populated up front by `newPreviewFor`, and the projection only later discovers it would be rejected, so the response asserted both `would_apply: false` and a status the transaction could never reach. A rejected projection has no resulting status — the real endpoint returns an error and writes no transaction — so `Finalize` clears it and the field is now `omitempty`, absent rather than empty. Applied at every projection exit and to bulk items, each of which carries its own status.

### Preview/real parity

Also added: tests asserting the preview and the real post **agree** on the failure paths — same HTTP status and same error code for a missing balance, and an identical body for a validation failure, which runs ahead of the `dry_run` branch.

These assert agreement rather than a literal code deliberately. The code a missing balance produces belongs to the error catalog and moves — #362 changes it from the generic not-found to a balance-specific one — so pinning the literal would either fight that change or quietly bless a divergence. Agreement is the property that matters, and it holds either way; both tests pass on this branch alone and with #362 merged.

### DIFF
--- a/api/model/transaction.go
+++ b/api/model/transaction.go
@@ -134,6 +134,14 @@ type BulkInflightVoidRequest struct {
 	// SkipQueue processes every item synchronously instead of routing them
 	// through the inflight-commit queue (the default).
 	SkipQueue bool `json:"skip_queue"`
+
+	// DryRun projects every void and returns the resulting balances without
+	// releasing any hold: no settlement is recorded and nothing is queued.
+	//
+	// Items are projected independently of one another, because that is how
+	// they really run — BulkInflightUpdate dispatches them across a worker
+	// pool with no ordering between them.
+	DryRun bool `json:"dry_run"`
 }
 
 // BulkInflightCommitItem describes one transaction in a bulk commit request.
@@ -154,6 +162,15 @@ type BulkInflightCommitRequest struct {
 	// SkipQueue processes every item synchronously instead of routing them
 	// through the inflight-commit queue (the default).
 	SkipQueue bool `json:"skip_queue"`
+
+	// DryRun projects every commit and returns the resulting balances without
+	// settling anything: no settlement is recorded, the holds stay in place,
+	// and nothing is queued.
+	//
+	// Items are projected independently of one another, because that is how
+	// they really run — BulkInflightUpdate dispatches them across a worker
+	// pool with no ordering between them.
+	DryRun bool `json:"dry_run"`
 }
 
 // BulkInflightResult is the per-item outcome reported in BulkInflightResponse.
--- a/api/model/transaction.go
+++ b/api/model/transaction.go
@@ -64,6 +64,15 @@ type BulkTransactionRequest struct {
 	Atomic       bool                 `json:"atomic"`
 	RunAsync     bool                 `json:"run_async"`
 	SkipQueue    bool                 `json:"skip_queue"`
+
+	// DryRun projects the batch and returns the result without applying any of
+	// it. The projection answers synchronously with 200, so run_async is
+	// ignored.
+	//
+	// Items are projected against each other's effects only when skip_queue is
+	// set, mirroring how the batch would really run; the response reports which
+	// mode was used.
+	DryRun bool `json:"dry_run"`
 }
 
 func (r *BulkTransactionRequest) ToBulkTransactionRequest() *model.BulkTransactionRequest {
@@ -90,6 +99,14 @@ type InflightUpdate struct {
 	// SkipQueue processes the commit/void synchronously instead of routing it
 	// through the inflight-commit queue (the default).
 	SkipQueue bool `json:"skip_queue"`
+
+	// DryRun projects the commit or void and returns the resulting balances
+	// without settling anything: no settlement transaction is recorded and the
+	// hold stays in place.
+	//
+	// A void always releases the full remaining hold, so any amount sent with
+	// one is ignored.
+	DryRun bool `json:"dry_run"`
 }
 
 // MaxBulkInflightItems caps the number of transactions accepted in a single
--- a/api/transaction_dryrun_bulk_api_test.go
+++ b/api/transaction_dryrun_bulk_api_test.go
@@ -1,0 +1,171 @@
+package api
+
+import (
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	model2 "github.com/blnkfinance/blnk/api/model"
+	"github.com/blnkfinance/blnk/internal/request"
+	"github.com/blnkfinance/blnk/model"
+)
+
+func bulkItem(source, destination string, amount float64, reference string) *model2.RecordTransaction {
+	return &model2.RecordTransaction{
+		Amount:      amount,
+		Precision:   100,
+		Currency:    "USD",
+		Source:      source,
+		Destination: destination,
+		Reference:   reference,
+		Description: "bulk dry run",
+	}
+}
+
+// TestDryRunBulkCumulativeCatchesIntraBatchShortfall covers the case the mode
+// exists for: with skip_queue the items really do run one after another, so the
+// second item must be judged against what the first one did.
+func TestDryRunBulkCumulativeCatchesIntraBatchShortfall(t *testing.T) {
+	router, b, err := setupRouter()
+	require.NoError(t, err)
+
+	source, destination := newDryRunFixture(t, b)
+
+	req := model2.BulkTransactionRequest{
+		DryRun:    true,
+		SkipQueue: true,
+		Transactions: []*model2.RecordTransaction{
+			// Drains almost all of the 500.00 the fixture funded.
+			bulkItem(source, destination, 450, "bulk_a_"+model.GenerateUUIDWithSuffix("ref")),
+			// Only affordable if the first item is ignored.
+			bulkItem(source, destination, 100, "bulk_b_"+model.GenerateUUIDWithSuffix("ref")),
+		},
+	}
+
+	body, err := request.ToJsonReq(&req)
+	require.NoError(t, err)
+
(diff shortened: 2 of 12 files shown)


Gee-Hive (commented):
Hi @Shivam8584 I noticed a few things on the PR:

using dry_run flag and we try to commit an amount that is greater than the orignal amount, 
we get a 200, with would_apply : true, this is wrong, kindly look into it

Gee-Hive (commented):
Also:

Inflight commit/void dry-run omits status and amount

POST /transactions and bulk with dry-run return both fields. Inflight dry-run does not:

`"dry_run": true,
  "would_apply": true,
  "operation": "commit",
  "status": "",
  "currency": "USD",
  "amount": 0,
  "precise_amount": "10000",
  "precision": 100`

precise_amount is correct; but status and amount are zero values.

Shivam8584:
Thanks @Gee-Hive — both points are fixed. Sorry the second one took a second pass; I addressed the first and didn't come back to it.

**1. `dry_run` commit above the original amount returned `would_apply: true`**

Real cause was in the handler, not the projection: the dry-run branch passed only `req.PreciseAmount` to `PreviewInflightAction`, skipping the `amount` → precision → `precise_amount` resolution the real commit path does a few lines below. With `precise_amount` unset that arrived as `nil`, and `nil`/zero is the sentinel for "commit the full remaining amount" — so an oversized `amount` was discarded rather than validated.

Moved the resolution above the `dry_run` check so both branches share it.

```
dry-run commit, amount 999999 against a 100.00 hold
→ 200 { "would_apply": false,
        "rejection": { "code": "TXN_COMMIT_AMOUNT_EXCEEDED",
                       "message": "cannot commit more than the original transaction amount. Original: USD100.00, Requested: USD999999.00" } }
```

Same code and message the real commit returns for that input.

**2. Inflight dry-run omitted `status` and `amount`**

`PreviewInflightAction` built its own `TransactionPreview` literal instead of going through `newPreviewFor`, and only set `Currency` and `Precision` as legs resolved — so both fields sat at their zero values, reading as though the projection had answered `""` and `0`.

`status` is now the status the settlement would carry once applied, matching the real endpoint (`APPLIED` for a commit, `VOID` for a void — not the internal `COMMIT`/`VOID` marker). `amount` is the float rendering of the precise total already reported, so the two always describe the same number.

Verified against your case, preview vs real, field for field:

```
dry-run commit:  status=APPLIED  amount=100  precise_amount=10000
real commit:     status=APPLIED  amount=100  precise_amount=10000

dry-run void:    status=VOID     amount=40   precise_amount=4000
real void:       status=VOID     amount=40   precise_amount=4000
```

---

Testing this also turned up seven further preview/reality mismatches, which are fixed on this branch — the largest being that a multi-leg preview only ever surfaced the *first* failing leg's reason, and that split previews ignored `skip_queue` and so projected cumulatively even on the queued path where legs run independently.

Separately, it surfaced bugs in existing code that are **not** part of this PR and are filed on their own: #360 (self-transfer always fails with a misleading lock error), #361 (negative amounts accepted via a `negative × negative` bypass), #362 (error-code accuracy), #363 (`GetBalanceAtTime` returns wrong historical balances on non-UTC servers), #364 (a state-dependent chain test).

I've also merged current `main` into this branch — CI was red on the NULL `parent_transaction` scan you fixed in #356, which this branch predated.

Gee-Hive (commented):
Hi @Shivam8584, thanks for attending to the comments. This PR is good to go.
Just one last edge. 

We always get {status: APPLIED} on {would_apply:false} transactions(POST/PUT)
This may look confusing/misleading to users. 

`  "dry_run": true,
    "would_apply": false,
    "rejection": {
        "code": "TXN_INSUFFICIENT_FUNDS",
        "reason": "insufficient_funds",
        "message": "failed to apply transaction to balances: insufficient funds in source balance"
    },
    "status": "APPLIED",`

Shivam8584:
Thanks @Gee-Hive — fixed and pushed.

You're right that it reads as contradictory: the response asserted both that the transaction would be rejected and that it would end up `APPLIED`.

**Cause**

`status` is populated up front, not at the end. `normalizePreviewStatus` sets the projected transaction to `APPLIED` (or `INFLIGHT`), `newPreviewFor` copies that onto the preview, and only later does the projection discover the balance can't cover it. At that point `would_apply` and `rejection` are set, but nothing revisits `status` — so it keeps describing an outcome that can't happen.

**Fix**

A rejected projection has no resulting status: the real endpoint returns an error and writes no transaction, so there is nothing to report. `status` is now cleared when `would_apply` is false, and the field is `omitempty`, so it's absent rather than empty — `would_apply` and `rejection` are the answer.

Your exact case now:

```json
{
  "dry_run": true,
  "would_apply": false,
  "rejection": {
    "code": "TXN_INSUFFICIENT_FUNDS",
    "reason": "insufficient_funds",
    "message": "failed to apply transaction to balances: insufficient funds in source balance"
  },
  "currency": "USD",
  "amount": 999999,
  "precise_amount": "99999900",
  "precision": 100,
  "balances": [ ... ]
}
```

No `status` key. The projected balances also correctly show `resulting_*` equal to `current_*`, since nothing would move.

An accepted projection is unchanged and still reports the status it would carry (`APPLIED`, or `INFLIGHT` for a hold).

Applied centrally at each projection exit rather than at the individual sites that mark a rejection, so the same holds for refunds, inflight commit/void, and per-item results inside a bulk projection — each of those carried its own `status` with the same problem.

**Tests**

Covered at both levels: unit tests on the projection itself, and an API-level regression test asserting the key is absent for an insufficient-funds dry run, plus one asserting the accepted path still reports `APPLIED`. I checked the API test fails without the fix, reproducing exactly the output you posted.

Gee-Hive (approved):

Issue #291, opened 2026-05-18
/refund-transaction cannot set meta_data/description on the reversal — it blind-inherits the original's


Shivam8584:
# Issue: `/refund-transaction` cannot set `meta_data` / `description` on the reversal — it blind-inherits the original's

**Type:** enhancement / API consistency
**Component:** `transaction.go` (`RefundTransaction`, `prepareRefundTransaction`), `api/transactions.go` (`RefundTransaction` handler)
**Version checked:** `main` @ `b37af21` (2026-05-16)

## Summary

When a transaction is refunded, the reversal transaction inherits the
**entire** `meta_data` and `description` of the original, and there is
no way for the caller to override either. A refund of a transaction
tagged `meta_data.type = "deposit"` is itself stored with
`meta_data.type = "deposit"` — so any system that classifies ledger
rows by metadata cannot distinguish a refund from the thing it reversed.

## Where it happens

`prepareRefundTransaction` (`transaction.go:2347`) builds the reversal
with a whole-struct copy:

```go
func prepareRefundTransaction(originalTxn *model.Transaction, skipQueue bool) *model.Transaction {
	newTransaction := *originalTxn          // copies meta_data, description, everything
	newTransaction.TransactionID = model.GenerateUUIDWithSuffix("txn")
	newTransaction.Reference     = model.GenerateUUIDWithSuffix("ref")
	newTransaction.ParentTransaction = originalTxn.TransactionID
	newTransaction.Source        = originalTxn.Destination  // swap
	newTransaction.Destination   = originalTxn.Source       // swap
	newTransaction.AllowOverdraft = true
	newTransaction.SkipQueue     = skipQueue
	...
}
```

Only id / reference / parent / source / destination / overdraft /
skipqueue are overridden. `MetaData` and `Description` ride along
unchanged, and nothing in the call chain
(`RefundTransaction` → `prepareRefundTransaction`, or the
`POST /refund-transaction/:id` handler) exposes a way to set them.

## Why it matters

`POST /transactions` already lets a caller set `meta_data` and
`description` at creation. A refund **is** a transaction — it is created
via the same `QueueTransaction` path — yet the refund endpoint is the
one transaction-creating path that cannot. That asymmetry forces a
caller who needs a correctly-classified refund to **stop using
`/refund-transaction` entirely** and hand-roll the reversal as a plain
`POST /transactions` with source/destination swapped — re-implementing
exactly what `prepareRefundTransaction` already does, just to get
control of two fields.

[...]

Issue #290, opened 2026-05-18
Proposal: first-class account statement endpoint (GET /balances/:id/statement)


Shivam8584:
## Summary

Add a first-class **account statement** endpoint, `GET /balances/:id/statement`, that returns a clean, ready-to-display ledger statement for a single balance — one row per *logical* money movement, annotated with direction (credit/debit) relative to that balance and a running balance.

Today, producing a statement requires every integrator to re-implement non-trivial, correctness-sensitive logic on top of `POST /transactions/filter`. The building blocks already exist in the codebase; this proposal wraps them into a single supported projection.

> Verified against `main` @ `e0f8676` (tag `v0.14.2`).

---

## Problem

Every transaction state transition is stored as a separate immutable row linked by `parent_transaction` (`QUEUED → INFLIGHT → APPLIED/VOID`, split/bulk parents, refunds). This is correct ledger design — but it means a naive transaction listing shows multiple rows per logical movement.

To build a wallet/account statement an integrator must currently:

1. **Collapse lifecycle rows** — exclude `QUEUED` parents and intermediate `INFLIGHT` holds so each logical movement appears once. The practical proxy is `status IN (APPLIED, VOID)`, but the API offers no endpoint that guarantees one row per logical transaction.
2. **Compute direction** — rows carry no `+/-`. The caller must compare `source`/`destination` against the balance ID for every row.
3. **Compute a running balance** — not derivable from the Filter API (no aggregation / window functions exposed; `BuildConditionExpression` only does flat `AND`/`OR`).
4. **Exclude lineage shadow transactions** — internal shadow rows share the `transactions` collection. Correct `balance_id` scoping hides them, but there is no explicit "exclude internal" flag.

Each consumer re-implements 1–4 inconsistently. Getting direction wrong means a statement with reversed signs. This is exactly the kind of projection the ledger should own once.

---

## Why this is feasible — the building blocks already exist

This is **not** a from-scratch feature. The hard parts are already in `main`:

- **Virtual `balance_id` field.** `internal/filter/sql_special.go::buildBalanceIdCondition` already expands `{"field":"balance_id","operator":"eq"}` into a single bracketed `(source = $n OR destination = $n)` condition for the `transactions` table — correctly `AND`-able with other filters. Per-balance scoping in one query is solved.
- **Statement-shaped index already exists.** `sql/...` defines:
  ```sql
  CREATE INDEX IF NOT EXISTS blnk_txn_created_at_not_queued_idx
[...]

PR #350, opened 2026-08-09
feat(transactions): let a refund set its own description and meta_data

Shivam8584:
Implements #291 — letting a refund set its own `description` and `meta_data` instead of blind-inheriting the original's.

**On the issue being closed:** you reviewed this back in May and said *"I'd say we go with option 1: api/transactions.go — extend refundTransactionRequest"*. It was then closed on 2026-08-06 in a sweep alongside #290, #225 and #345 — but unlike #345 (fixed in `c8c003c`) nothing had landed for this one, and `refundTransactionRequest` still carries only `skip_queue` on `main`. Reading that as housekeeping rather than a decision, so here's the implementation. Happy to close this if you'd rather not take it.

**Stacked on #348** — please merge that first; this branch contains its three commits, so the only new commit here is `c36a753`. Both PRs touch `refundTransactionRequest` and `prepareRefundTransaction`, so they can't be independent. Rebases to `main` cleanly once #348 lands, or I can rebase this to go first if you prefer.

### What changes

```jsonc
POST /refund-transaction/:id
{
  "description": "refund for ticket #4821",
  "meta_data": { "type": "refund", "reason": "goodwill" }
}
```

A refund of a transaction tagged `meta_data.type = "deposit"` was itself stored as a `"deposit"`, so anything classifying ledger rows by metadata couldn't tell a refund apart from the movement it reversed.

### Metadata is merged, not replaced

Worth correcting something in my own issue: I justified merging by saying a replace would drop `QUEUED_PARENT_TRANSACTION`. That's wrong — that key is set in `QueueTransaction` (`transaction_queue.go:102`), which runs *after* `prepareRefundTransaction`, so it isn't at risk.

Merging is still right, for different keys. `setTransactionMetadata` writes `inflight`, `atomic` and `allow_overdraft` onto the original, and those are inherited by the struct copy and read downstream (`transaction_inflight.go:83`, `transaction_coalescing.go:604`, `transaction_rejection.go:71`). A blind replace would drop them. Merging lets a caller reclassify the reversal while they survive.

It also gives callers a handle on a live wart: `setTransactionMetadata` only ever *sets* the inflight marker and never clears it, so a reversal of an inflight transaction inherits `inflight: true` even though `prepareRefundTransaction` explicitly sets `Inflight = false` — which `transformTransaction` then reports back (that's #293 surfacing here). Passing `{"meta_data": {"inflight": false}}` corrects it on that row. Not fixing #293 in this PR.

### One bug worth calling out

The merge builds a **new** map rather than writing into the inherited one. `newTransaction := *originalTxn` shares the original's metadata map, so writing through it would have rewritten **the transaction being refunded**. There's a test for exactly this (`TestPrepareRefundDoesNotMutateOriginalMetaData`), and I verified against a live server that the original row still reads `type: deposit` after a refund overrides it to `type: refund`.

### Threading is additive

Following how `skip_queue` was added to this endpoint in `8c2022a`:

| | |
|---|---|
| `RefundTransaction` | signature unchanged; delegates to `RefundTransactionWithOptions` |
| `RefundWorkerWithOptions` | signature unchanged; `RefundWorkerWithRefundOptions` added alongside |
| `RefundWorker` | untouched — atomic bulk rollback still uses it |
| `prepareRefundTransaction` | unexported, takes `RefundOptions` (2 call sites) |

`RefundTransaction` is exported and called from 13 test sites, and isn't behind an interface, so changing its signature would be source-breaking for SDK consumers. Nothing existing changes, and an absent body or absent fields reproduces today's behaviour exactly.

### Tests

10 new tests: description override, metadata merge with caller keys winning and ledger keys surviving, the aliasing guard above, overriding an inherited `inflight`, empty overrides being ignored rather than blanking inherited values, `skip_queue` behaviour preserved, and voided-original status handling unchanged.

Verified live: overriding produced `{"type": "refund", "reason": "goodwill", "channel": "card", "allow_overdraft": true}` — caller keys won, uninvolved keys inherited, ledger keys survived — while the original row was untouched in Postgres. A bodiless `POST` still returns 201 and inherits both fields.

### DIFF
--- a/api/model/transaction.go
+++ b/api/model/transaction.go
@@ -44,6 +44,16 @@ type RecordTransaction struct {
 	Destinations       []model.Distribution   `json:"destinations"`
 	MetaData           map[string]interface{} `json:"meta_data"`
 	EffectiveDate      *time.Time             `json:"effective_date,omitempty"`
+
+	// DryRun projects the transaction's effect on the source and destination
+	// balances and returns the result without applying it. No transaction row
+	// is written, no balance changes, nothing is queued, no webhook fires, and
+	// the reference is not consumed.
+	//
+	// A dry run always responds synchronously with 200 rather than 201, and
+	// takes precedence over skip_queue: a projection is inherently an immediate
+	// answer about what a real post would do.
+	DryRun bool `json:"dry_run"`
 }
 
 // BulkTransactionRequest is the public API request shape for creating a batch
--- a/api/refund_options_api_test.go
+++ b/api/refund_options_api_test.go
@@ -1,0 +1,103 @@
+package api
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/blnkfinance/blnk/internal/request"
+	"github.com/blnkfinance/blnk/model"
+)
+
+// TestRefundAppliesDescriptionAndMetaData covers the endpoint contract: the
+// reversal can be described and classified as its own movement rather than
+// inheriting the original's.
+func TestRefundAppliesDescriptionAndMetaData(t *testing.T) {
+	router, b, err := setupRouter()
+	require.NoError(t, err)
+
+	source, destination := newDryRunFixture(t, b)
+
+	applied, err := b.QueueTransaction(t.Context(), &model.Transaction{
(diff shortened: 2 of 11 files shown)


Gee-Hive (approved):
Hi @Shivam8584 ,I have tested and this PR looks good

Thank you

PR #351, opened 2026-08-09
fix(database): include meta_data in GetBalanceByIDLite

### LINKED ISSUE #256: Balance metadata on typesense is empty after recording a transaction with the balance.
**Describe the bug**
When a balance performs a transaction, the Typesense upsert does not include the metadata field. As a result, when the balance is reindexed after the transaction, the existing metadata in Typesense is overwritten and becomes empty.

**To Reproduce**
1. Create two new balances:
```
curl -X POST 'http://localhost:5001/balances' \
-d '{
  "ledger_id": "general_ledger_id",
  "currency": "USD",
  "meta_data": {
    "field": "test_value"
  }
}'
```

2. Call search to confirm that the balances are indexed with the metadata
```
curl -X POST 'http://localhost:5001/search/balances' \
-d '{ "q": "*" }'
```

3. Create a transaction between both balances. 
```
curl -X POST 'http://localhost:5001/transactions
-d '{
    "amount": 12091.12,
    "currency": "USD",
    "precision": 100,
    "reference": "ref-001",
    "source": "bln_19c0bcd8-e136-4a89-91b3-5be4066b68c4",
    "destination": "bln_a603ab29-7f3d-40f9-9752-8f04e197ddd9",
    "description": "Bug test",
    "allow_overdraft": true
}'
```

4. Check the balances via Search again.
5. You'll see that the balances' metadata is now empty/overwritten although other fields are correct. 

**Expected behavior**
The metadata should persist when updating a balance record in Typesense after creating a transaction.

vjymisal0:
## Bug

`GetBalanceByIDLite` (`database/balance.go`) selects `balance_id, indicator, currency, ledger_id, balance, credit_balance, debit_balance, inflight_balance, inflight_credit_balance, inflight_debit_balance, created_at, version, track_fund_lineage, allocation_strategy, identity_id` — `meta_data` is not in the list, so every `*model.Balance` it returns has `MetaData == nil`, no matter what's stored in Postgres.

This function is the default path in `getSourceAndDestination` (`transaction_execution.go`) used to look up the source/destination balance for a transaction when `EnableQueuedChecks` is off. `postTransactionActions` then hands that balance straight to `search.NewIndexBatch` / `batch.AddDependency("balances", ...)`, which re-upserts it into Typesense. Because `MetaData` is nil on the struct (not because it was actually cleared in the DB), the balance's real metadata gets wiped from the search index on every transaction touching that balance.

Repro from the issue:
1. Create a balance with `meta_data`.
2. Confirm it's indexed correctly via `/search/balances`.
3. Run any transaction against that balance.
4. Search again — `meta_data` is now empty, even though `SELECT meta_data FROM blnk.balances WHERE balance_id = ...` still shows it in Postgres.

## Fix

Add `meta_data` to the SELECT and scan it into a `[]byte`, then unmarshal into `balance.MetaData` when non-empty — the same nil-safe pattern already used by `GetSourceDestination` and `GetAllBalances` elsewhere in this file, so this brings `GetBalanceByIDLite` in line with its siblings rather than introducing a new pattern.

## Scope note

`GetBalancesByIDsLite` (the bulk version, used by `transaction_coalescing.go` and `lineage_allocation.go`) has the identical gap — its SELECT also omits `meta_data`. I left it untouched since issue #256 is specifically about the single-lookup path and I wanted to keep this PR focused; happy to open a follow-up for the bulk path if useful.

## Testing

- Updated the three existing `GetBalanceByIDLite` sqlmock tests for the new query text/columns.
- Added `TestGetBalanceByIDLite_Success` assertion that `meta_data` round-trips (was previously not asserted at all, which is how this shipped).
- Added `TestGetBalanceByIDLite_NilMetaData` for the NULL `meta_data` case.
- `go test ./database/...` passes (three env-dependent `_RealDB` tests skip locally, no Postgres running — unrelated to this change).
- `go vet ./...` and `golangci-lint run ./database/...` are clean on the changed files (one pre-existing, unrelated staticcheck note elsewhere in `balance.go` at line ~1455, outside this diff).
- `gofmt` clean.

Fixes #256

### DIFF
--- a/account_test.go
+++ b/account_test.go
@@ -53,8 +53,8 @@ func TestCreateAccount(t *testing.T) {
 	}
 	metaDataJSON, _ := json.Marshal(account.MetaData)
 
-	rows := sqlmock.NewRows([]string{"balance_id", "indicator", "currency", "ledger_id", "balance", "credit_balance", "debit_balance", "inflight_balance", "inflight_credit_balance", "inflight_debit_balance", "created_at", "version", "track_fund_lineage", "allocation_strategy", "identity_id"}).
-		AddRow(account.BalanceID, "", "NGN", account.LedgerID, 100, 50, 50, 0, 0, 0, time.Now(), 0, false, "FIFO", "")
+	rows := sqlmock.NewRows([]string{"balance_id", "indicator", "currency", "ledger_id", "balance", "credit_balance", "debit_balance", "inflight_balance", "inflight_credit_balance", "inflight_debit_balance", "created_at", "version", "track_fund_lineage", "allocation_strategy", "identity_id", "meta_data"}).
+		AddRow(account.BalanceID, "", "NGN", account.LedgerID, 100, 50, 50, 0, 0, 0, time.Now(), 0, false, "FIFO", "", nil)
 
 	mock.ExpectQuery("SELECT .* FROM blnk.balances WHERE balance_id =").
 		WithArgs(account.BalanceID).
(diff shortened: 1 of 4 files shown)


jerry-enebeli:
Hi @vjymisal0, thanks for the PR. There are some failing tests. Could you please look into them?

PR #353, opened 2026-08-10
feat(balances): allow creating General Ledger balances with an explicit indicator

Gee-Hive:
General Ledger balances can now be created with an explicit indicator on POST /balances, so @ accounts (e.g. @cash) can be provisioned up front instead of only appearing on first use.

indicator is optional, GL-only (ledger_id = general_ledger_id), must look like @name, and duplicates for the same (indicator, currency) return 409 instead of an empty 201.

Test plan

 Create a GL balance with indicator: "@cash" — succeeds

 Reject bad indicators (no @, whitespace, non-GL ledger)

 Creating the same indicator + currency again returns 409

 Existing create-balance paths (no indicator) still works

### DIFF
--- a/balance.go
+++ b/balance.go
@@ -18,8 +18,8 @@ package blnk
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/blnkfinance/blnk/config"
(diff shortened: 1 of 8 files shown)


balance.go
```diff
@@ -231,11 +232,23 @@ func (l *Blnk) CreateBalance(ctx context.Context, balance model.Balance) (model.
...
 	balance, err := l.datasource.CreateBalance(balance)
 	if err != nil {
 		span.RecordError(err)
 		return model.Balance{}, err
 	}
+
+	// The datasource signals a duplicate (indicator, currency) pair by returning
+	// an empty balance with a nil error (see database/balance.go unique_indicator_currency
+	// handling). Without this check the API would respond 201 with an empty body.
+	if requestedIndicator != "" && balance.BalanceID == "" {
```
jerry-enebeli: I think the error could be handled by checking the error code vs empty id/indicator

jerry-enebeli (approved):

PR #355, opened 2026-08-17
fix(lock): release locks on a context detached from cancellation

### LINKED ISSUE #354: Balance locks are never released when the request context is cancelled
**Describe the bug**
Every redis lock release goes through `Locker.Unlock`, which runs a Lua `Eval` on the
caller's context

```go
// internal/lock/lock.go:107
func (l *Locker) Unlock(ctx context.Context) error {
	script := "if redis.call('get', KEYS[1]) == ARGV[1] then return redis.call('del',
          KEYS[1]) else return 0 end"
	result, err := l.client.Eval(ctx, script, []string{l.key}, l.value).Result()
	...
}
```

**To Reproduce**
1. Start blnk with default config (`lock_duration` unset → 5m as of `config/config.go:54`).
2. Issue a transaction against a balance and abandon the request before it completes so the 
   client-side timeout shorter than the transaction.
3. Inspect redis: the lock key for that balance is still present with a live TTL.
4. Issue any further transaction touching the same balance → `failed to acquire lock`.
5. `DEL` the key by hand → service is restored immediately.

**Expected behavior**
Releasing a lock should not depend on the liveness of the request that took it. A cancelled
context is the normal case for cleanup, not an exceptional one. `Unlock` should release on a
context detached from cancellation with its own short timeout, e.g.

```go
releaseCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)
defer cancel()
```

so cleanup survives cancellation while still being bounded.

**Screenshots**

**Platform (please complete the following information):**
 - Blnk 0.14.5, also reproduced against 0.15.2; code above is current `main` (30d35fb).
 - Postgres + redis via docker compose; Go 1.25.

**Additional context**

toniastro:
Locker.Unlock ran its redis Eval on the caller's context, so a cancelled or timed-out request left the key to expire on its own TTL and every later transaction on that balance failed with "failed to acquire lock". Nothing surfaced it: rollback discards the error, releaseLock and releaseSingleLock only log it.

Fixes #354

### DIFF
--- a/internal/lock/lock.go
+++ b/internal/lock/lock.go
@@ -100,11 +100,15 @@ func (l *Locker) Lock(ctx context.Context, timeout time.Duration) error {
 
 // Unlock releases the lock if the calling instance is the lock holder (based on the value).
 // The operation is atomic, ensuring only the holder of the lock can release it.
+// ctx's cancellation is ignored: an abandoned request must still free the lock.
 // Parameters:
 // - ctx: The context for managing the unlock request lifecycle.
 // Returns an error if the unlock operation fails, either because the lock expired or
 // the caller is not the lock holder.
 func (l *Locker) Unlock(ctx context.Context) error {
+	ctx, cancel := releaseContext(ctx)
+	defer cancel()
+
 	// Lua script ensures atomicity: checks the value and deletes the key if the value matches.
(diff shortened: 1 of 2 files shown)


jerry-enebeli:
Thank you @toniastro

Issue #334, opened 2026-07-19
Feature: opt-in to allow zero-amount transactions (fixed-shape multi-leg postings)


sinkovsky:
## Summary

Blnk rejects any transaction whose `precise_amount` is `0`:

```
POST /transactions  { "precise_amount": 0, ... }
→ 400 {
  "error": "failed to apply transaction to balances: transaction precise amount must be positive",
  "error_detail": { "code": "TXN_VALIDATION_ERROR" }
}
```

Requesting an opt-in way (a `blnk.json` config flag, or a per-request `allow_zero_amount`) to permit zero-amount transactions. They would be no-ops on balances but persist as normal records (hash-chain, references, `meta_data`).

## Use case

We post a class of operations that always has the same fixed set of legs, where one leg can legitimately net to zero for a given instance (e.g. an internal customer/house split adjustment that is sometimes 0). Because the 0-amount leg is rejected, we must conditionally omit it — so the same logical operation is sometimes N legs and sometimes N−1. A fixed leg shape simplifies downstream reconciliation/auditing that keys off structure, and bulk-atomic batches where every leg is expected.

## Current behavior

Validation rejects `precise_amount <= 0` with `TXN_VALIDATION_ERROR` ("transaction precise amount must be positive"). Verified on **v0.15.1**.

## Proposal

- Add an opt-in flag (`transaction.allow_zero_amount` in `blnk.json`, or a per-request `allow_zero_amount: true`).
- Default **off** — preserves current behavior. Zero-amount transactions apply as no-ops but record like any other.

Happy to contribute a PR if this is acceptable — what is the preferred location for the guard (per-request field vs global config)?

jerry-enebeli:
Hi @sinkovsky, let me think through this and get back to you. But i see your point.

Issue #137, opened 2025-05-08
Don't save zero amount transactions


stkegoul:
**Describe the bug**
Transactions with $0 amount get recorded in the ledger.

## To Reproduce
1. Set up a transaction with the following structure
```
{
  "amount": 1,
  "reference": "[[${guid}]]",
  "currency": "USD",
  "precision": 100,
  "sources": [
    {
      "identifier": "@Balance2",
      "distribution": "left"
    },
    {
      "identifier": "@Balance1",
      "distribution": "1"
    }
  ],
  "destination": "@EntryFeePool",
  "description": "Fund with starting balance amount",
  "allow_overdraft": false,
  "inflight": true,
  "skip_queue": true
}
```

2. Ensure @Balance1 has enough funds to cover the fee (e.g., $1).
3. Execute the transaction

## Expected Behavior
1. The fee should be fully deducted from @Balance1 since it has sufficient funds.
2. No $0 transaction should be recorded for @Balance2.

## Actual Behavior
Both transactions are recorded, including a $0 transaction for @Balance2, along with in-flight and empty transactions.

PR #361, opened 2026-08-20
fix(transactions): reject negative amount and precision at validation

Shivam8584:
## Summary

A transaction with a negative `amount` is **accepted and applied** when `precision` is also negative, because precision multiplies:

```
amount(-25) × precision(-100) = precise_amount(+2500)
```

The product is positive, so every downstream positivity check passes and the transaction applies normally — but `amount: -25` and `precision: -100` are **persisted as sent**.

```bash
curl -X POST /transactions -d '{
  "amount": -25, "precision": -100,
  "currency": "USD", "source": "bln_A", "destination": "bln_B",
  "reference": "...", "description": "...",
  "allow_overdraft": true, "skip_queue": true
}'
# -> HTTP 201, status: APPLIED, balances move
```

## Impact

**Balances stay correct.** Double-entry holds, insufficient-funds is still enforced, and no money is created or destroyed — I verified this on every surface, including through refund, void, and bulk-commit round-trips. This is not a fund-theft bug and I don't want to overstate it.

**The damage is to the record**, and to anything downstream that reads `amount`. In a ledger of four such transactions:

```
SUM of reported `amount`       = -105
SUM of actual `precise_amount` = 10500   (= +105.00)
```

A reporting or reconciliation job summing `amount` sees the sign inverted. And these records are invisible to positive-amount queries:

```
POST /transactions/filter  {"field":"amount","operator":"gt","value":0}  -> 0 rows
POST /transactions/filter  {"field":"amount","operator":"lt","value":0}  -> 9 rows
```

A routine "all transactions where amount > 0" report silently returns **none** of them, despite each having moved real money. `precise_amount / precision` also recomputes to `-25`, so the record is internally consistent in its wrongness.

## Why nothing caught it

`model.Transaction.validate()` already intends to enforce this:

```go
func (transaction *Transaction) validate() error {
	if transaction.PreciseAmount != nil {
		if transaction.PreciseAmount.Sign() <= 0 {
			return errors.New("transaction precise amount must be positive")
		}
		return nil          // <-- early return
	}
	if transaction.Amount <= 0 {          // <-- unreachable in practice
		return errors.New("transaction amount must be positive")
	}
	return nil
}
```

`PreciseAmount` is set on every path that reaches this, so the non-nil branch always returns early after checking only the (positive) product — the `Amount` check never runs. That left the effective guard against a negative `amount` to be an *unrelated split-distribution check* firing by accident on a negative total (`validateDistributions`, where `fixedTotalDec (0) > totalAmountDec (negative)` is true), which the sign flip bypasses.

Positive amounts are clearly the intended policy — the codebase says so in its own error text, `validateRequestedAmount` enforces it on the inflight-commit path, and #334 requests zero-amount as an *opt-in* precisely because it's currently rejected.

## Fix

Three parts:

1. Reject negative `amount` / `precise_amount` and negative `precision` in `RecordTransaction.ValidateRecordTransaction` — before anything is queued or persisted, covering every creation surface through one validator (bulk items included).
2. Check `Amount` unconditionally in `model.Transaction.validate()`, ahead of the `PreciseAmount` early-return, so the existing deep guard works as defense in depth.
3. Reject negative precision on `MonitorCondition`, which accepted it too.

**Zero is deliberately left alone** in both places: `precision: 0` means "not supplied" and is defaulted to 1 downstream (`model.ApplyPrecision`), and zero-amount handling has existing semantics (#137, #334) this change should not alter.

## Test plan

- [x] 7 new cases in `TestValidateRecordTransaction` covering the bypass, each negative field alone, and the two must-not-break zero cases
- [x] **All 12 surfaces that previously accepted the bypass now return 400**: plain (both queue modes), inflight, scheduled, `precise_amount` + negative precision, bulk, atomic bulk, unbounded scale (`-1 × -1e9`), negative amount with normal precision, negative `precise_amount`, split fan-in, and balance monitors
- [x] **8 legitimate-traffic regression cases still pass**: normal transaction, `precision: 0`, `precise_amount`-only, inflight, split fan-out, bulk, balance monitor, scheduled
- [x] Zero negative-amount records reach the ledger after the full matrix
- [x] `go test -short ./api/... ./model/... .` passes
- [x] Full `go test -short ./...` — `cmd`, `database`, `internal/request`, `internal/search` fail identically without this change (TypeSense/network-sandbox/balance-snapshot flakiness, unrelated to validation)

Found while adversarially testing the dry-run preview PR (#349) against the real API; unrelated to that PR and lives in existing core validation.

### DIFF
--- a/api/model/model.go
+++ b/api/model/model.go
@@ -45,6 +45,63 @@ func validatePrecisionIsInteger(value interface{}) error {
 	return nil
 }
 
+// validatePrecisionNotNegative rejects a negative precision.
+//
+// Precision is a scaling factor — "the number you used to convert the
+// smallest unit to an integer" — so it is only meaningful as a positive
+// value. Zero is left valid because it means "not supplied" and is
+// defaulted to 1 downstream (see model.ApplyPrecision).
+//
+// A negative precision is not merely meaningless, it is load-bearing for
+// the bypass described on validateAmountNotNegative below: it flips the
+// sign of the computed precise amount.
+func validatePrecisionNotNegative(value interface{}) error {
+	precision, ok := value.(float64)
+	if !ok {
+		return errors.New("invalid precision type")
+	}
+
+	if precision < 0 {
+		return errors.New("precision cannot be negative")
+	}
+
+	return nil
+}
+
+// validateAmountNotNegative rejects negative amounts on both the float
+// `amount` field and the big.Int `precise_amount` field.
+//
+// The ledger's intent is that amounts are positive — model.Transaction's own
+// validate() reports "transaction amount must be positive" — but that check
+// runs deep in the apply path and, for `amount`, is unreachable: by the time
+// it runs, PreciseAmount has been computed and its non-nil branch returns
+// early. That left the effective guard against a negative `amount` to be an
+// unrelated split-distribution check firing by accident on a negative total.
+//
+// That accidental guard is bypassable, because precision multiplies:
+//
+//	amount(-25) x precision(-100) = precise_amount(+2500)
+//
+// The product is positive, so every downstream positivity check passes, and
+// the transaction applies — while `amount: -25` and `precision: -100` are
+// persisted as sent. The stored record then reports the opposite sign of the
+// movement it describes, and is invisible to an `amount > 0` filter.
+//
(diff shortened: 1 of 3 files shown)


jerry-enebeli:
looks good, thanks @Shivam8584

PR #362, opened 2026-08-20
fix(errors): report the resource that was actually not found, and stop leaking internals

Shivam8584:
## Summary

Five accuracy problems in error responses, found while auditing the API surface. Each is scoped against the published contract at [docs.blnkfinance.com/advanced/error-codes](https://docs.blnkfinance.com/advanced/error-codes), which states that `error_detail.code` is **stable and is what clients should branch on**, while `error` / `error_detail.message` are *"for display and may change between releases."*

Four of the five are message-only or correct a response that contradicted the docs. **The fifth adds a new code and is deliberately a separate, droppable commit** — see below.

## 1. A missing *balance* was reported as `TXN_NOT_FOUND`

```
POST /transactions  (source = a balance id that does not exist)
→ 404  { "code": "TXN_NOT_FOUND", "message": "Balance with ID 'bln_...' not found" }
```

The code said transaction-not-found; the message said balance. A client branching on the code went hunting for a missing transaction when a balance was missing.

`database/balance.go` raised the generic `apierror.ErrNotFound`, and the transaction handlers apply `withUpgrade(ErrGenNotFound → ErrTxnNotFound)`, which blanket-upgrades *any* generic not-found — including a balance lookup performed while processing a transaction.

This contradicted the docs, which already document **`BAL_NOT_FOUND` — "Balance does not exist"**. Now raised at the source (plus `BAL_MONITOR_NOT_FOUND` for monitors). The balance endpoints already mapped these via `withUpgrade`, so **their responses are unchanged**; only the mislabelled transaction-route case moves, onto its documented code.

## 2. Bulk item indexes disagreed between layers

```
validation layer:  "transactions[0]"                  + details.index = 0   ← 0-based
queueing layer:    "failed to queue transaction 1"                          ← 1-based
```

Same batch, same item, two different numbers depending on which layer rejected it — an off-by-one for any client using it to highlight the offending row. Both are 0-based now. Message text only; the structured `details.index` is untouched.

## 3. Refunding a non-refundable transaction said "no transaction to refund"

`GetRefundableTransactionsByParentID` filters on existence **and** refundable status in one query, so an empty result cannot distinguish "no such transaction" from "exists but is not refundable" — but the message implied only the former, sending you to look for a wrong/deleted id when the transaction was sitting right there in `INFLIGHT`.

It now states both causes and names the id. The code stays `TXN_NOT_FOUND`, which the docs already define as covering a *"refundable transaction"* that was not found.

## 4. Request-decoding errors leaked Go internals

```
POST /transactions  { "precise_amount": "5000", ... }
→ 400  "math/big: cannot unmarshal \"\\\"5000\\\"\" into a *big.Int"
```

The Go type name is meaningless to an API client and exposes an implementation detail. That shape is rewritten to name the field and expected format; anything unrecognised passes through unchanged, so this narrows a leak without hiding decode failures.

## 5. `"not found in DB or queue"` named internal storage

"DB or queue" describes how Blnk stores a pending transaction internally — not something a client can act on, and it constrains the implementation's freedom to change where a transaction lives before it settles. The message still classifies to `TXN_NOT_FOUND` via the existing pattern; code unchanged.

---

## ⚠️ The last commit is different in kind — drop it if you disagree

`a5f7346` adds **`TXN_ALREADY_REFUNDED`**. Three of the four "already in that state" conditions get a domain-specific code; the refund one was the odd one out:

```
"Transaction already committed" → TXN_ALREADY_COMMITTED
"has already been voided"       → TXN_ALREADY_VOIDED
"has already been used"         → TXN_DUPLICATE_REFERENCE
"has already been refunded"     → GEN_CONFLICT          ← generic
```

Unlike the others, **this is not correcting a contradiction.** `GEN_CONFLICT` is documented as *"Request conflicts with current resource state"*, which an already-refunded transaction genuinely is — so today's behaviour is not *wrong*, just less useful than its siblings. `TXN_ALREADY_REFUNDED` appears in neither the catalog nor the docs, so adopting it is a taxonomy decision for you, and would need a new row in the published error-code reference.

Worth noting: `api/errors_test.go` **explicitly pinned** `GEN_CONFLICT` for this case, sitting between siblings that pin their specific codes. I've updated that assertion — but if that pin was deliberate rather than descriptive, this commit is the one to drop. It is self-contained.

## Test plan

- [x] Missing balance on a transaction route → `BAL_NOT_FOUND`
- [x] **Regression:** missing transaction still → `TXN_NOT_FOUND`; `GET /balances/:id` still → `BAL_NOT_FOUND`
- [x] Both bulk layers report `transactions[0]`
- [x] Refund on an `INFLIGHT` transaction names the id and both causes
- [x] `precise_amount` as a string no longer mentions `big.Int`; ordinary malformed JSON still reported
- [x] Refund on an unknown id no longer says "DB or queue"
- [x] Already-refunded → `TXN_ALREADY_REFUNDED`; **regression:** already-voided still → `TXN_ALREADY_VOIDED`
- [x] `go test -short ./api/... ./model/... ./internal/apierror/... .` passes
- [x] Full `go test -short ./...` — `cmd`, `database`, `internal/request`, `internal/search` fail identically without this change (TypeSense / network-sandbox / balance-snapshot flakiness, unrelated)

Found while adversarially testing the dry-run preview PR (#349) against the real API; unrelated to that PR.

### DIFF
--- a/api/errors.go
+++ b/api/errors.go
@@ -34,7 +34,13 @@ import (
 // the legacy field clients depend on today (flat "error"/"errors" string,
 // preserved verbatim) and the structured "error_detail" object with the
 // canonical error code. The legacy field is removed — and error_detail
-// renamed to error — at the next major release. See docs/errors.md.
+// renamed to error — at the next major release.
+//
+// The code catalog and its stability guarantee are published at
+// https://docs.blnkfinance.com/advanced/error-codes (source:
+// blnkfinance/blnk-docs, advanced/error-codes.mdx). error_detail.code is the
+// stable field clients branch on; error, errors and error_detail.message are
+// for display and may change between releases.
 
 const errorDetailKey = "error_detail"
 
--- a/api/errors.go
+++ b/api/errors.go
@@ -42,6 +42,26 @@ const errorDetailKey = "error_detail"
 // 5xx responses so database/driver details never leak to clients.
 const sanitizedInternalMessage = "internal server error"
 
+// sanitizeBindError rewrites a request-decoding error into a message about
+// the request, rather than about Blnk's internals.
+//
+// encoding/json and math/big report decode failures in terms of Go types.
+// A precise_amount sent as a JSON string, for example, surfaces verbatim as:
+//
+//	math/big: cannot unmarshal "\"5000\"" into a *big.Int
+//
+// The Go type name means nothing to an API client and exposes an
+// implementation detail, so the known shape is rewritten to name the field
+// and the expected format instead. Anything unrecognised is returned
+// unchanged — this narrows a leak, it does not hide decode failures.
+func sanitizeBindError(err error) string {
+	msg := err.Error()
+	if strings.Contains(msg, "big.Int") || strings.HasPrefix(msg, "math/big:") {
+		return "invalid numeric value: precise amount fields must be sent as a JSON number in the smallest unit (for example 5000, not \"5000\")"
+	}
+	return msg
+}
+
 type respondOptions struct {
 	defaultCode     apierror.ErrorCode
 	fallbackMessage string
--- a/api/errors.go
+++ b/api/errors.go
@@ -218,7 +218,7 @@ var messagePatterns = []messagePattern{
 	{[]string{"not in inflight status"}, apierror.ErrTxnNotInflight},
 	{[]string{"Transaction already committed"}, apierror.ErrTxnAlreadyCommitted},
 	{[]string{"has already been voided"}, apierror.ErrTxnAlreadyVoided},
-	{[]string{"has already been refunded"}, apierror.ErrGenConflict},
+	{[]string{"has already been refunded"}, apierror.ErrTxnAlreadyRefunded},
 	{[]string{"has already been used"}, apierror.ErrTxnDuplicateReference},
 	{[]string{"cannot commit more than"}, apierror.ErrTxnCommitAmountExceeded},
 	{[]string{"insufficient funds"}, apierror.ErrTxnInsufficientFunds},
--- a/api/errors.go
+++ b/api/errors.go
@@ -228,6 +228,15 @@ var messagePatterns = []messagePattern{
 	{[]string{"has already been used"}, apierror.ErrTxnDuplicateReference},
 	{[]string{"cannot commit more than"}, apierror.ErrTxnCommitAmountExceeded},
 	{[]string{"insufficient funds"}, apierror.ErrTxnInsufficientFunds},
+	// UpdateBalances reports a settlement that outruns its hold as
+	// "insufficient inflight debit balance" / "insufficient inflight credit
+	// balance", which the pattern above does not match. Unclassified, the
+	// condition inherited whichever fallback the calling handler happened to
+	// set — GEN_BAD_REQUEST from the inflight route, GEN_INTERNAL elsewhere —
+	// so the same failure surfaced under different codes depending on where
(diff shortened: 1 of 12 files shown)


Gee-Hive (commented):
Hi @Shivam8584 , thanks for this PR. I have a few feedbacks

1. Bind error sanitization — trim the message
Good fix for the math/big leak on bulk/inflight binds. Please help drop the inline format example (for example 5000, not "5000") from the sanitized message — docs should cover JSON shape; the API only needs to say the value is invalid and which field class failed (e.g. precise amount must be a JSON number).

2. Bulk missing balance — still TXN_NOT_FOUND
Single-txn create correctly returns BAL_NOT_FOUND for a missing source balance. But bulk txn still returns TXN_NOT_FOUND, not the BAL_NOT_FOUND for a missing source balance. Please fix

3. Refund not-found —  revert to it's previous response
Please let's revert the empty refund result message to "no transaction to refund" (code stays TXN_NOT_FOUND). The new text is too verbose for an error response; txn status details belong on GET /transactions/:id and in docs, not in the 404 body.

The BAL_NOT_FOUND at source, bulk index alignment, and removal of “DB or queue” from refund internals all look good to merge once the above are addressed.

Shivam8584:
Thanks @Gee-Hive — all three addressed and pushed.

1. **Bind message trimmed** to `invalid numeric value: precise amount must be a JSON number`.
2. **Bulk now returns `BAL_NOT_FOUND`.** Cause was upstream of the API: `CreateBulkTransactions` flattened the failure to `errors.New(...)`, so the typed error was gone and the code had to be guessed from a sentence that matches `{"transaction","not found"}` first. It now returns a `BulkTransactionError` that keeps the item's error underneath, and the handler resolves the code through a new `resolveErrorCode`. `result.Error` and `batch_id` unchanged. Covered by `TestBulkMissingBalanceReportsBalanceNotFound`, which fails on the old code with your exact symptom. (Applies to `skip_queue: true`; the queued path acknowledges before any balance is read.)
3. **Refund not-found reverted** to `"no transaction to refund"`.

I also applied `resolveErrorCode` to `hookFailureCode` and `GetReconciliation`, which had the same defect. No response changes today — their only typed not-found already carries a 404.

**One thing to decide before merging:** this changes exactly one status. An internal database failure inside a sync bulk batch now returns **500 instead of 400** (the old path couldn't see the typed error and fell through to `GEN_BAD_REQUEST`). Every other case — insufficient funds, duplicate reference, invalid amount, `sql.ErrNoRows`, unclassifiable — resolves exactly as before. I think 500 is correct, but if you'd rather keep 400 for everything that isn't a not-found, say so and I'll push the two-line restriction.

Shivam8584:
One follow-up, from checking whether the bulk fix left the same defect anywhere else.

It did, in two places: `hookFailureCode` and `GetReconciliation` both asked `classifyMessage(err.Error())` whether a failure was a not-found, while holding the error itself. A not-found raised as a typed `APIError`, or as a wrapped `sql.ErrNoRows`, is invisible to that check unless its wording happens to match a pattern — and the request is then answered as an infrastructure failure (500 / HOOK_OPERATION_FAILED) instead of a 404. Both now go through `resolveErrorCode`.

I went through the remaining `classifyMessage` call sites: `respondError` and `resolveErrorCode` already try the chain first, and `respondPreview` classifies a projection's rejection *message*, which has no error behind it — that one is correct as it stands.

Covered by `TestHookFailureCodeUsesTheErrorChain`.

Gee-Hive (approved):

PR #363, opened 2026-08-20
fix(balances): put the snapshot bound and the caller's cutoff on the same clock as the columns they bound

Shivam8584:
> **Rewritten after review.** The original version of this PR changed `balance_snapshots.snapshot_time` from `TIMESTAMPTZ` to `TIMESTAMP` and pinned the snapshot writer's clock. @Gee-Hive asked to keep the column as-is rather than rewrite the table. Both migrations are now gone — **this PR contains no migration and no schema change.** History and the discussion below are left intact.

## Summary

Two independent off-by-one-offset bugs in `GetBalanceAtTime`, both fixed in Go, both silent — no error is raised, the endpoint just returns a wrong historical balance.

**1. The snapshot bound loses the server's offset.** `blnk.balance_snapshots.snapshot_time` is `TIMESTAMPTZ`, so the driver returns it in the session's zone. `getMostRecentSnapshot` hands that value straight to `fetchTransactions` as the lower bound of a range over `created_at` / `effective_date`, which are plain `TIMESTAMP`:

```sql
WHERE COALESCE(effective_date, created_at) > $snapshot_time
```

A `TIMESTAMP` comparison *discards* the offset rather than applying it, so the bound sits one UTC offset away from the transaction timeline — those columns hold UTC wall clocks, because every writer normalises with `txn.CreatedAt.UTC()` before the insert. East of UTC the range narrows and post-snapshot transactions are dropped; west of UTC it widens and pre-snapshot transactions are applied on top of the snapshot that already accounts for them.

**2. The caller's cutoff loses the client's offset.** `GET /balances/:id/at` parses `?timestamp=` with `time.Parse(time.RFC3339)`, keeping whatever offset the client wrote, and the same `TIMESTAMP` comparison discards it. This one is wrong **on a UTC server too** — it depends on the caller's offset, not the deployment's. On `main` at session TZ `UTC`, `?timestamp=…+05:30` returns 1500 where the identical instant written `+00:00` returns 1000.

## The fix

Two conversions, at the two points where a value crosses from an offset-carrying representation into an offset-discarding comparison:

- `snapshotTime.UTC()` where `getMostRecentSnapshot` returns the bound
- `targetTime.UTC()` on entry to `GetBalanceAtTime`

No SQL, no migration, no table rewrite.

## Why no migration is needed

`snapshot_time` already stores the correct **instant**. The writer's `v_current_time TIMESTAMP := NOW()` round-trips through the session zone losslessly on the way into a `TIMESTAMPTZ` column, so `main`'s writer was never the problem — only the read path was. Pinning the writer to `(NOW() AT TIME ZONE 'UTC')` while leaving the column as `TIMESTAMPTZ` actively does *not* work: Postgres reads that UTC wall clock back as session-local, which fixes east of UTC and leaves west of UTC untouched.

Measured end to end, snapshot taken by the real function, correct credit 400:

| session TZ | `main` | writer pin only, no type migration | this PR |
|---|---|---|---|
| `UTC` | 400 | 400 | 400 |
| `Asia/Kolkata` | **0** | 400 | 400 |
| `America/New_York` | **1400** | **1400** | 400 |

## Why this went unnoticed

The shipped `docker-compose.yaml` pins `TZ: ${TZ:-Etc/UTC}` and CI runs UTC, so bug 1 is mathematically invisible in the reference deployment. Bug 2 is not — it is reachable from any client that sends a non-UTC offset, on any deployment.

## Tests

`TestGetBalanceAtTime_NonUTCSession_RealDB` pins bug 1 across session zones `UTC`, `Asia/Kolkata` and `America/New_York`, with the snapshot written by the production function rather than a hand-written row, on a `MaxOpenConns(1)` pool so the session `TimeZone` actually sticks. `TestGetBalanceAtTime_TargetTimeOffset_RealDB` pins bug 2 across three client offsets. Neither needs a CI change.

Two existing fixtures inserted `created_at` raw while production inserts `txn.CreatedAt.UTC()`, which made unrelated tests depend on the developer's laptop zone; they now write the way production writes.

`TestGetMostRecentSnapshot_Found` compared the returned time with `assert.Equal`, which compares the zone as well as the instant. It now asserts the instant with `.Equal()` and pins the location separately.

## Test plan

- [x] Fresh migration from an empty database applies **37** migrations — same as `main` — and `snapshot_time` is still `timestamp with time zone`
- [x] Full `./database/` suite green at host TZ `Etc/UTC`, `Asia/Kolkata` and `America/New_York`
- [x] Each variant in the table above measured against its own migrated database, not inferred
- [x] `go build ./...` and `go vet ./database/` clean

Found while auditing test failures during work on #349; unrelated to that PR.

### Known residual, not fixed here

On a non-UTC session the writer's `NOW()` → `TIMESTAMP` → `TIMESTAMPTZ` round trip is ambiguous during the DST fall-back hour, so a snapshot taken in that one hour a year can be stored up to an hour off. Only the column type change closes that, and it isn't worth the table rewrite.

### Suggested follow-up (not in this PR)

Running CI in one non-UTC time zone would have caught bug 1 for free, and would guard the rest of the ledger's time arithmetic against the same class of bug.


Shivam8584:
Extra evidence while auditing on a non-UTC machine: this is not a latent bug — an **existing test in the repo already fails** without this migration.

On stock `main`, against a fresh database, on a host in `IST (+0530)`:

```
--- FAIL: TestGetBalanceAtTime_RealDB
    Messages: the newer snapshot (-90m) must win over the older one (-2h)
```

Applying this PR's migration to the same tree and database:

```
--- PASS: TestGetBalanceAtTime_RealDB
ok  github.com/blnkfinance/blnk/database
```

`TestGetBalanceAtTime_RealDB` is not added here — it is already on `main`. It passes in CI only because the workflow and the shipped compose both pin UTC, where the `TIMESTAMPTZ`/`TIMESTAMP` offset is zero. On any server with a non-zero offset, snapshot selection picks the wrong snapshot and the existing test catches it.

That is also an argument for running one CI job in a non-UTC timezone: the coverage for this already exists, it just never runs anywhere it can fail.

Gee-Hive (commented):
Hi @Shivam8584, thanks for the clear migration notes on the mixed-type comparison issue.

But I’m not sure what this change buys us much in practice.

Comparing TIMESTAMPTZ to plain TIMESTAMP isn't the real issue. Even with the same type on both sides, if session timezone isn't UTC, we can still produce wrong historical balances. This change doesn't address that. This PR only removes Postgres's extra coercion on top of that 

On our docs standard path (docker-compose.yaml with TZ=Etc/UTC), session timezone is already UTC, so these comparisons behave correctly as-is. This PR doesn’t change how values are written: Go still stores transactions as UTC-naive timestamps, snapshots still use SQL NOW() in session time.

Bottom line: For guaranteed correctness, run with UTC (TZ=Etc/UTC on the Blnk process and Postgres session, which is what docker-compose defaults to). the PR  removes the extra Postgres coercion when TIMESTAMPTZ is compared to TIMESTAMP. It does not remove the need for UTC — non-UTC can still produce wrong snapshot-based historical balances.

A few things I’d like your take on:

Please what exactly does the type change fix? Moving snapshot_time from TIMESTAMPTZ to plain TIMESTAMP removes Postgres session-timezone coercion during comparison — but it doesn’t change how values are written (Go still stores UTC-naive txns; snapshots still use SQL NOW()). Beyond aligning column types, what does this save us that UTC enforcement alone doesn’t?


What’s the worst case if we leave it as it is? Can you please walk through a realistic Blnk deployment (not just a misconfigured local DB) where the mixed types cause wrong GetBalanceAtTime results, and confirm this migration fully fixes that case?



Im trying to understand whether this is fixing an active gap in our documented deployment model, or its hardening against setups we don’t support.

Shivam8584:
Thanks @Gee-Hive — you were right on both counts, and my earlier evidence was wrong. `TestGetBalanceAtTime_RealDB` fails on a non-UTC host only because its fixture inserts `created_at` raw while production inserts `txn.CreatedAt.UTC()`. The host's zone does not matter, as you said.

The real gap is the **Postgres session** TimeZone, and the type change alone does not close it — it swaps one wrong answer for another. Measured end to end, snapshot taken by the real function, correct credit 400:

| session TZ | stock main | type change only | type + writer |
|---|---|---|---|
| `UTC` | 400 | 400 | 400 |
| `Asia/Kolkata` | **0** | **1400** | 400 |
| `America/New_York` | **1400** | **1400** | 400 |

Note it is not only double-counting: east of UTC the range narrows and post-snapshot transactions are silently dropped.

Three commits now: the column type; `v_current_time := (NOW() AT TIME ZONE 'UTC')` in the snapshot writer; and `targetTime.UTC()` at the datasource, because `?timestamp=` keeps the caller's offset and a `TIMESTAMP` comparison discards it — that one is wrong on `main` today even at session TZ `UTC` (`+05:30` returns 1500 where `+00:00` returns 1000). Plus the fixtures, so the suite stops depending on the developer's laptop. `TestGetBalanceAtTime_NonUTCSession_RealDB` pins all of it and needs no CI change.

**Two things for you:**

1. **Operational:** `1781200000` rewrites the table under ACCESS EXCLUSIVE — 2.55s per million rows measured. Worth a release note if `balance_snapshots` is large.
2. **Scope:** if you'd rather answer this with a documented UTC requirement, the third commit still stands on its own (it is not timezone-configuration dependent) and I'll split it out. Say which and I'll push.

Verified on the upgrade path, which CI never exercises: from a `main`-migrated database, `migrate up` applies exactly 2 and re-runs apply 0; every instant survives including both DST boundaries and across a Down/Up round trip; indexes and the immutability trigger survive; full suite green against an upgraded-in-place database and under three session zones; `-race` clean.

Shivam8584:
One more, found while checking that the two changes above actually closed the whole thing rather than most of it.

There is a **third** clock in the same comparison: the timestamp the caller supplies. `GET /balances/:id/at` parses `?timestamp=` with `time.Parse(time.RFC3339)`, which keeps whatever offset the client wrote. That value then bounds two ranges over plain `TIMESTAMP` columns — and a `TIMESTAMP` comparison **discards** an offset rather than applying it.

Measured on the existing code, one transaction inside the cutoff (1000) and one outside it (500), asking for the same instant three ways:

| `?timestamp=` offset | credit |
|---|---|
| `+00:00` | 1000 |
| `+05:30` | **1500** |
| `-04:00` | **0** |

So `...T15:00:00+05:30` cuts off at 15:00 UTC — five and a half hours past the instant asked for — and quietly includes everything in the gap.

This one is **not** introduced by the snapshot work; `created_at` has always been compared this way, on `main`, at any session TimeZone including UTC. But making `snapshot_time` a plain `TIMESTAMP` brings snapshot selection under the same rule (it was instant-compared before, being `TIMESTAMPTZ`), so shipping the first two changes without this one would have widened an existing bug rather than fixed anything. That is the only reason it is in this PR and not a separate one — say the word and I'll split it out.

Fixed by normalising once at the datasource entry point, which is the single place both ranges are bounded from. `TestGetBalanceAtTime_TargetTimeOffset_RealDB` asserts the same instant written in three zones returns the same balance; it fails on `main` with exactly the numbers above.

That accounts for all three writers into the comparison — transactions (already UTC), snapshots (now UTC), and the caller's cutoff (now UTC) — so `GetBalanceAtTime` no longer depends on any timezone anywhere. I also checked there is no fourth: after `1781200000` the only remaining `TIMESTAMPTZ` columns are `blnk.api_keys` and `gorp_migrations`, neither of which is compared against a transaction timestamp.

Shivam8584:
Last one — I went looking for what I'd missed rather than assuming there was nothing, and there was.

Running the **whole** `database` suite across every combination of host zone and Postgres session zone turned up failures I hadn't seen, because I'd only been running the targeted tests at non-UTC. The cause wasn't the migrations: the time-travel fixtures insert `created_at` / `effective_date` / `snapshot_time` straight from `time.Now()`, keeping the **host's** zone, while production normalises to UTC before the insert. Once the caller's target time is normalised (previous commit), those host-local fixture rows no longer line up with it.

So the fixtures were the last thing in the path still recording a local wall clock — which is also why `TestGetBalanceAtTime_RealDB` fails on an IST laptop on `main` today, for reasons unrelated to what it asserts. That was the basis of my very first comment on this PR, and it was measuring the fixture, not the code.

Normalised the three fixture inserts. The suite is now independent of both clocks:

| | pg=UTC | pg=America/New_York | pg=Asia/Kolkata |
|---|---|---|---|
| **host=UTC** | ok | ok | ok |
| **host=Asia/Kolkata** | ok | ok | ok |
| **host=America/New_York** | ok | ok | ok |

Nine for nine; the middle and bottom rows failed before this commit.

I also confirmed against a freshly migrated database that there is exactly one snapshot writer — `take_daily_balance_snapshots_batched`, no overloads — that its body carries the UTC pin, that both `snapshot_time` and `created_at` are `timestamp without time zone`, and that `getMostRecentSnapshot` is the only reader of `snapshot_time` anywhere in the codebase.

That's the whole path accounted for: three writers, one reader, one caller-supplied bound.

Gee-Hive (commented):
Thanks @Shivam8584 for responding. PR looks good, happy to merge. 

One ask: can we drop the schema type migration (Remove the ALTER COLUMN migration + Revert the snapshot_time type change (keep TIMESTAMPTZ))?

On our standard UTC deployment the type mismatch doesn’t bite — TIMESTAMPTZ vs TIMESTAMP comparisons behave correctly as long as the Postgres session is UTC. We’d rather avoid rewriting balance_snapshots in production. The writer UTC pin, targetTime.UTC(), and the test fixes are what we want.

One note: TestGetBalanceAtTime_NonUTCSession_RealDB was written assuming the type migration lands — without it, the non-UTC subtests won’t pass. Fine to trim that test to UTC-only (or drop it) since we’re documenting UTC as a requirement.

Shivam8584:
Done — dropped, and the writer migration had to go with it. Both are gone, `snapshot_time` stays `TIMESTAMPTZ`, and there is now **no migration in this PR at all**.

The reason the writer pin couldn't stay: with the column left as `TIMESTAMPTZ`, `(NOW() AT TIME ZONE 'UTC')` produces a UTC wall clock that Postgres then reads back as *session-local* on the way into the column. That fixes east of UTC and leaves west of UTC exactly as wrong as `main`. Measured, snapshot written by the real function, correct credit 400:

| session TZ | `main` | your ask (writer pin, no type migration) | this push |
|---|---|---|---|
| `UTC` | 400 | 400 | 400 |
| `Asia/Kolkata` | **0** | 400 | 400 |
| `America/New_York` | **1400** | **1400** | 400 |

So shipping the writer pin alone would have left a migration in the tree whose 40 lines of rationale claim a correctness property it doesn't deliver.

**What replaces it is one line in Go.** `snapshot_time` already stores the right *instant* — `NOW()` round-trips through the session zone losslessly on the way in, so `main`'s writer was never the problem. The offset is lost on the way **out**: the driver returns the value in the session's zone, and `getMostRecentSnapshot` passes it straight to `fetchTransactions` as the lower bound of a range over `created_at` / `effective_date`, which are plain `TIMESTAMP`. That comparison *discards* the offset instead of applying it, so the bound lands one UTC offset away from the transaction timeline. `snapshotTime.UTC()` at the one point the value crosses from a `TIMESTAMPTZ` read into a `TIMESTAMP` comparison puts it back on the same clock as the columns it bounds.

That means the non-UTC test doesn't need trimming — `TestGetBalanceAtTime_NonUTCSession_RealDB` passes at all three session zones as written, so the coverage survives even though the schema doesn't change. Happy to trim it to UTC-only anyway if you'd still rather the suite not assert anything about non-UTC sessions; say the word.

`targetTime.UTC()` and the fixture changes are untouched — that half was never timezone-configuration dependent (`+05:30` still returns 1500 where `+00:00` returns 1000 on `main` at session TZ `UTC`).

One test change worth flagging: `TestGetMostRecentSnapshot_Found` asserted the returned time with `assert.Equal`, which compares the zone as well as the instant, so normalising to UTC tripped it. It now asserts the instant with `.Equal()` and pins the location separately.

Verified against a database migrated from this branch — 37 migrations, same as `main`, `snapshot_time` still `timestamp with time zone`. Full `./database/` suite green at host TZ `Etc/UTC`, `Asia/Kolkata` and `America/New_York`, minus `TestConnectDB_Success` / `TestGetDBConnection_Singleton`, which fail the same way on stock `main` here because they dial a hardcoded `localhost:5432`.

Residual worth knowing, not fixed here: on a non-UTC session the writer's `NOW()` → `TIMESTAMP` → `TIMESTAMPTZ` round trip is ambiguous during the DST fall-back hour, so a snapshot taken in that one hour a year can be stored up to an hour off. Only the type change closes that, and it isn't worth the rewrite.

Gee-Hive (approved):

Issue #183, opened 2025-08-20
Atomic bulk transactions fail when payload contains multiple sources or multiple destinations


adamprescott:
**Describe the bug**
When performing an atomic bulk transaction, if one of the transactions has `"allow_overdraft": false` and it's below a transaction that has `"allow_overdraft": true`, the destination balance is still updated with the transaction with an overdraft.

**Current work around:** Put the transactions without overdraft facilities at the top of the array.

**To Reproduce**
POST to /transactions/bulk

**Example Payload**
```json
{
    "atomic": true,
    "inflight": false,
    "run_async": false,
    "skip_queue": true,
    "transactions": [
        {
            "amount": 100,
            "reference": "res_adam-testing-clientBal-01 - 907522cb-93ad-45bb-939b-c9e208366580 - deposit-pm",
            "currency": "GBP",
            "precision": 1,
            "sources": [
                {
                    "distribution": "100",
                    "identifier": "bln_19d666bd-ee2a-4d38-a391-793983004d27"
                }
            ],
            "destination": "bln_39c49732-39c7-4548-9398-1c2c2b7a8dd5",
            "description": "Reservation deposit (payment methods) for reference res_adam-testing-clientBal-01",
            "allow_overdraft": true
        },
        {
            "amount": 300,
            "reference": "res_adam-testing-clientBal-01 - 907522cb-93ad-45bb-939b-c9e208366580 - deposit-client",
            "currency": "GBP",
            "precision": 1,
            "source": "bln_52893611-7222-453e-9291-fad7e5c08d58",
            "destination": "bln_39c49732-39c7-4548-9398-1c2c2b7a8dd5",
            "description": "Reservation deposit (client balance) for reference res_adam-testing-clientBal-01",
            "allow_overdraft": false
        }
    ]
}
```
> [!NOTE]
> The payment methods transaction is covering multiple Fiat currency sources e.g. Cash, Visa, Mastercard, etc... hence it's an array of sources. The Client Balance is money held on account on behalf of a customer, they're able to purchase services via this account, it can be paid into with things like Gift Cards, etc...

**Sample Response**
```json
{
    "batch_id": "bulk_fe0ebaf0-b0c3-4a6c-8e26-287238b02ae4",
[...]

PR #368, opened 2026-08-21
fix(transactions): reject splitting sources and destinations at once

Shivam8584:
## Summary

A transaction carrying **both** `sources[]` and `destinations[]` is accepted, then fails downstream with a message about neither:

```bash
POST /transactions
{
  "amount": 100, "precision": 100, "currency": "USD",
  "sources":      [{"identifier":"bln_a","distribution":"60"}, {"identifier":"bln_b","distribution":"40"}],
  "destinations": [{"identifier":"bln_c","distribution":"60"}, {"identifier":"bln_d","distribution":"40"}]
}

→ 404 TXN_NOT_FOUND   "Balance with ID '' not found"
```

## This combination is already documented as invalid

Both documentation pages state the restriction explicitly, and the second states it twice:

> **[Multiple sources](https://docs.blnkfinance.com/transactions/multiple-sources):** "When sending from multiple sources, you can only send to one destination."

> **[Multiple destinations](https://docs.blnkfinance.com/transactions/multiple-destinations):** "When sending to multiple destinations, you can only send from one source."

So this PR does not introduce a new rule — it enforces one the docs already commit to, and which the API has been accepting anyway.

## Why the failure looks unrelated to the cause

`SplitTransactionPrecise` distributes over **one side only** — `Sources` when present, `Destinations` otherwise — and clears both on the legs it produces:

```go
var ds []Distribution
if len(transaction.Sources) > 0 {
    ds = transaction.Sources
} else if len(transaction.Destinations) > 0 {
    ds = transaction.Destinations
}
```

A transaction carrying both therefore has its `Destinations` **silently dropped**, and every leg is left with an empty `Destination`. The blank id in that error is not something the caller sent — it is what the split left behind.

## The status is wrong too, not just the message

`404` reads as *"not there yet"*. A client with retry-on-not-found logic will retry a request that can never succeed, because the real fault is a malformed shape. This should be a `400` the caller never retries.

## Fix

Reject the combination while the shape is still visible, naming the actual conflict rather than a field that only *appears* missing as a consequence of it:

```
400 TXN_VALIDATION_ERROR
source: a transaction can split either sources or destinations, not both
```

The check sits alongside the existing `sourceOrSourcesValidation` / `destinationOrDestinationsValidation` rules, which already reject the singular-plus-plural combinations on each side individually. This closes the remaining one: plural on **both** sides.

## Test plan

- [x] New `TestSplitOnOneSideValidation` covering both-sides-split (rejected) and the three legitimate shapes — fan-in, fan-out, plain — accepted
- [x] Verified live: both-sides now returns the same `400` and identical message on the real and preview paths; fan-in, fan-out and plain transactions are unaffected
- [x] Full `go test -short ./...` passes (Postgres, Redis, TypeSense running)
- [x] `golangci-lint` clean

### Notes

Related: #183 ("Atomic bulk transactions fail when payload contains multiple sources or multiple destinations") covered a different failure for payloads of this shape. This PR addresses the validation gap that lets the shape through in the first place.

This also removes a preview/reality divergence for callers of the `dry_run` work in #349: that endpoint reported `400 TXN_VALIDATION_ERROR "destination is required"` for this input while the real post reported `404 TXN_NOT_FOUND` — the preview was right about the category and wrong about the reason, since the caller did supply destinations. Validation runs ahead of both branches, so the two now agree by construction, on a message that describes the actual mistake. The fix stands on its own regardless of #349.

### DIFF
--- a/api/model/model.go
+++ b/api/model/model.go
@@ -63,6 +63,30 @@ func destinationOrDestinationsValidation(t *RecordTransaction) validation.RuleFu
 	}
 }
 
+// splitOnOneSideValidation rejects a transaction that splits both sides at
+// once.
+//
+// SplitTransactionPrecise distributes over one side only — it takes Sources
+// when they are present and Destinations otherwise — and clears both on the
+// legs it produces. A transaction carrying both therefore has its
+// Destinations silently dropped, and every leg is left with an empty
+// Destination, which surfaces downstream as a lookup for a balance with a
+// blank id:
+//
+//	Balance with ID '' not found
+//
+// That names neither the field at fault nor the reason, for a request the
+// ledger cannot express in the first place. Reject it where the shape is
+// still visible.
+func splitOnOneSideValidation(t *RecordTransaction) validation.RuleFunc {
+	return func(value interface{}) error {
+		if len(t.Sources) > 0 && len(t.Destinations) > 0 {
+			return errors.New("a transaction can split either sources or destinations, not both")
+		}
+		return nil
+	}
+}
+
 func (l *CreateLedger) ValidateCreateLedger() error {
 	return validation.ValidateStruct(l,
 		validation.Field(&l.Name, validation.Required),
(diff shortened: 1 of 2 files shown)


Gee-Hive (approved):

Gee-Hive (approved):
Thanks @Shivam8584 for this PR, Please kindly help resolve merge conflict

Shivam8584:
Thanks @Gee-Hive — conflict resolved and pushed.

Both sides had added a validator to `ValidateRecordTransaction` in the same place: this branch's `splitOnOneSideValidation` and main's `sameBalanceValidation` from #374. They check different things, so both are kept and both stay wired — split-on-one-side on `Source`, same-balance on `Destination`.

Merged `main` in rather than rebasing so the review history stays attached. The net diff against `main` is unchanged from what you approved: `api/model/model.go` +26/-1 and `api/model/model_test.go` +87. Ready to merge.

PR #372, opened 2026-08-21
test: run the backup happy path against a pg_dump stub instead of skipping it

Shivam8584:
## Summary

`TestBackupDB/Successful_backup_to_disk` was gated behind an environment variable that nothing sets, so the backup happy path had never executed — in CI or anywhere else:

```go
if os.Getenv("BLNK_TEST_PG_DUMP") == "" {
	t.Skip("BLNK_TEST_PG_DUMP not set; skipping pg_dump-dependent test")
}
```

```console
$ grep -rn 'BLNK_TEST_PG_DUMP' --include=*.go --include=*.yml --include=*.yaml .
api/admin_api_test.go:47:   if os.Getenv("BLNK_TEST_PG_DUMP") == "" {
api/admin_api_test.go:48:       t.Skip("BLNK_TEST_PG_DUMP not set; ...")
```

Three matches, all of them the gate itself.

## The fix

Replace the flag with an executable `pg_dump` **stub** placed first on `PATH`, reusing the convention `internal/pg-backups` already uses for exactly this.

Depending on a *real* `pg_dump` would only half-fix it. PostgreSQL is explicit:

> "pg_dump cannot dump from PostgreSQL servers newer than its own major version; it will refuse to even try, rather than risk making an invalid dump." — [pg_dump documentation](https://www.postgresql.org/docs/current/app-pgdump.html)

CI runs `postgres:latest` against `ubuntu-latest`'s bundled client, so a real-binary gate would still skip on the machine where it matters most. A stub removes the environment from the question: what this test covers is the endpoint wiring — route, config, `BackupDir`, response — and that is worth asserting on every run rather than never.

The test also now asserts a **non-empty `.sql` file actually landed** under `BackupDir`. A 200 alone would be returned even if `pg_dump` wrote nothing.

Verified it passes with no `pg_dump` on `PATH` at all. The deployed image is unaffected: the Dockerfile installs `postgresql-client-16` and compose pins `postgres:16`.

## Scope

This PR originally also un-skipped `TestSlackNotification_HangingReceiverShouldTimeOut`, which needed an adjustable timeout on the shared `request.Call` client. @Gee-Hive was right to push back on that, and on reflection the replacement I tried wasn't worth it either: the assertions that survived without a global knob amounted to comparing a constant against a literal, and proving `Call` uses the package-level client. Neither justifies moving production code for a test's benefit.

So that half is gone. `internal/request/request.go` and the notification test are byte-identical to `main`, the timeout stays fixed at 30s, and that test stays skipped — asserting the real bound costs a 30s wall-clock wait, and that trade-off was correct as originally judged.

**Net diff: one file, `api/admin_api_test.go`.**


Gee-Hive (commented):
Hi @Shivam8584, thanks for the PR:
The backup test fix looks good.

One concern: request.Call is shared — Slack and account-number generation both use it. With SetCallTimeout, anyone can change the timeout for the whole process, not just the test. 

This can cause logical interference: while the timeout is changed other callers of request.Call would also use that timeout.

There’s also no validation — the setter accepts 0 (disables the timeout) and negative values (fail immediately).

Changing production code just to make one test faster isn’t a safe approach. Please let's Keep the timeout fixed at 30s, and test it in a way that doesn’t need a global knob.

Shivam8584:
Thanks @Gee-Hive — you're right, and the knob is gone. Pushed.

Both of your objections were correct: `request.Call` is shared, so a process-wide setter lets any caller reshape Slack's and account-number generation's timeouts as a side effect, and it accepted `0` (no bound at all) and negatives (fail immediately). Validating it would have made it a safer bad idea, not a good one.

### What the timeout looks like now

Fixed at 30s, and unreachable from outside the package:

```go
const callTimeout = 30 * time.Second

var callClient = &http.Client{Timeout: callTimeout}
```

`Call` uses `callClient` instead of building a client per call. That was worth doing on its own — a fresh `http.Client` each time throws away the connection pool — and it leaves exactly one place the bound is set.

### Testing it without a knob

The property moved to `internal/request`, where the client can be substituted from inside the package. Two tests, both running, together ~0.2s:

- `TestCallClientTimeoutIsFixed` — the bound exists and is 30s.
- `TestCallReturnsWhenTimeoutFires` — the bound is applied *to the request*. A client whose `Timeout` is set but never used would pass the first test and hang here. Against a receiver that accepts the connection and never answers, `Call` must return a `net.Error` with `Timeout() == true`. It runs against a short-lived substitute client, restored on return; the substitution is package-local, tests in the package don't run in parallel, and no caller can reach it.

### The Slack-level test

Removed rather than rewritten. Without a knob, asserting it there means a 30s wall-clock wait, and the bound it was checking isn't that package's to begin with — it belongs to `request.Call`, which is where it's now asserted. A comment in the file points at the two tests above so the next reader doesn't have to rediscover this. `TestSlackNotification_ConnectionRefused` still covers the local-failure side.

Net effect on the PR's original goal: the two tests that never ran now run — the backup happy path on every run, and the timeout property in 0.2s instead of never — with no production code changed for the sake of a test.

Shivam8584:
@Gee-Hive you were right that changing production code to make one test faster isn't safe, so I cut that half rather than argue for it.

`internal/request/request.go` and `notification_dispatch_test.go` are now byte-identical to `main`, the timeout stays fixed at 30s, and `TestSlackNotification_HangingReceiverShouldTimeOut` stays skipped — verifying the real bound costs a 30s wall-clock wait, and that call was right as originally made.

**What's left is one file, `api/admin_api_test.go`.** `BLNK_TEST_PG_DUMP` is set by nothing in the repo, so that test had never run anywhere; it now runs against a `pg_dump` stub and asserts a non-empty `.sql` actually landed under `BackupDir` (a 200 comes back even if `pg_dump` writes nothing). Retitled and rewrote the description to match.

Happy to close this instead if a single test isn't worth the review time — no objection either way.

Issue #367, opened 2026-08-21
Auth middleware rejects POST /{id}/metadata for scoped keys as unknown resource type


stkegoul:
## Summary

When secure mode is on, `POST /{id}/metadata` fails for **scoped API keys** with `403 Forbidden: Unknown resource type`, even if the key has `metadata:write` or `*:*`.

The metadata handler already supports balance, ledger, transaction, and identity IDs. Auth never reaches it.

The **master key still works**. Secure mode off also works.

Reported against Core **v0.15.1**. The same path-mapping logic is still on `main`.

Docs: [Update Metadata](https://docs.blnkfinance.com/reference/update-metadata)

## Reproduction

1. Enable secure mode.
2. Create a scoped key with `metadata:write` (or `*:*`).
3. Call:

```http
POST /bln_<balance-id>/metadata
X-Blnk-Key: <scoped-key>
Content-Type: application/json

{
  "meta_data": {
    "example": "value"
  }
}
```

The same 403 happens for ledger, transaction, and identity IDs (`ldg_…`, `txn_…`, `idt_…`).

## Expected

Auth should treat `POST /{id}/metadata` as a metadata write and check `metadata:write`. A valid scoped key should reach the handler.

## Actual

```text
403 Forbidden: Unknown resource type
```

(`AUTH_UNKNOWN_RESOURCE`)

## Why

The route is:

```go
router.POST("/:entity-id/metadata", a.UpdateMetadata)
```

`getResourceFromPath` only reads the **first** path segment and maps it through `pathToResource` (`ledgers`, `balances`, `metadata`, …). It already has special cases for `mocked-account`, `multi-search`, and `refund-transaction`, but not for this route.

For `POST /bln_xxx/metadata`:

| Segment | Value |
|---|---|
| First | `bln_xxx` (entity ID, not a resource name) |
| Last | `metadata` (the actual resource) |

`bln_xxx` is unknown, so auth aborts **before** checking scopes. That is why even `*:*` fails.

Master key is not affected: it is accepted and returned before path mapping runs.

## Fix direction

In `getResourceFromPath`, recognise `/:entity-id/metadata` (for example, if the last segment is `metadata`) and return `ResourceMetadata`, matching the existing special cases.

Cover with auth tests for:

- scoped key with `metadata:write` on `POST /bln_…/metadata`, `POST /ldg_…/metadata`, `POST /txn_…/metadata`, `POST /idt_…/metadata` → 200 (auth allowed)
[...]

PR #376, opened 2026-08-24
fix(auth): map /:id/metadata for scoped API keys

Gee-Hive:
Auth only inspected the first path segment, so POST /{entity-id}/metadata failed with AUTH_UNKNOWN_RESOURCE before metadata:write was checked.


JohnAkindipe:
Good work @Gee-Hive, might it be useful to add explicit tests for "*:*" (wildcard) scoped keys as well for completeness sake.

JohnAkindipe:
Since this PR addresses issue #367, that issue should be closed, no?

PR #387, opened 2026-09-04
effective_date dropped in some cases where skip_queue = false

TheFifthRider:
# Problem
There are cases when a correctly populated effective_date on a queued transaction will not be populated on the subsequent applied transaction, incorrectly ignoring a backdate that should have been applied to a transaction.

Specifically:
* Getting stuck queued transactions
* Getting queued transactions for coalescing

## Recommended solution
Add effective_date to the queries and properly populate the transaction model

## Background
When backfilling blnk from another ledger, I ran into an issue where ~70%+ of my transactions incorrectly were applied without an effective_date, causing blnk to treat them as though they had occurred on the date the backfill ran. After some troubleshooting, it appears the issue was that in certain cases when queueing transactions, blnk will fail to query for the queued transaction's effective_date field, and will instead persist the applied transaction without the effective_date field.

Digging into the database, I could see that the queued transaction records were all being recorded with the correct effective_date, but the subsequent applied transaction would not.

When running the backfill again with skip_queue=true, all applied transactions had an effective_date.

### DIFF
--- a/database/transaction_coalescing.go
+++ b/database/transaction_coalescing.go
@@ -36,7 +36,7 @@ func (d Datasource) GetQueuedTransactionsForCoalescing(ctx context.Context, sour
 	defer span.End()
 
 	rows, err := d.Conn.QueryContext(ctx, `
-		SELECT transaction_id, parent_transaction, source, reference, amount, precise_amount, precision, currency, destination, description, status, created_at, meta_data, scheduled_for, hash
+		SELECT transaction_id, parent_transaction, source, reference, amount, precise_amount, precision, currency, destination, description, status, created_at, meta_data, scheduled_for, hash, effective_date
 		FROM blnk.transactions t
 		WHERE t.status = 'QUEUED'
 		  AND t.source = $1
@@ -75,7 +75,7 @@ func (d Datasource) GetQueuedTransactionsForSourceCoalescing(ctx context.Context
 	defer span.End()
 
 	rows, err := d.Conn.QueryContext(ctx, `
-		SELECT transaction_id, parent_transaction, source, reference, amount, precise_amount, precision, currency, destination, description, status, created_at, meta_data, scheduled_for, hash
+		SELECT transaction_id, parent_transaction, source, reference, amount, precise_amount, precision, currency, destination, description, status, created_at, meta_data, scheduled_for, hash, effective_date
 		FROM blnk.transactions t
 		WHERE t.status = 'QUEUED'
 		  AND t.source = $1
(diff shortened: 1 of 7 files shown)


Gee-Hive (commented):
Thanks @TheFifthRider  for this PR — the coalescing/stuck-queue fix looks good, I was able to reproduce the bug and confirm the APPLIED rows keep effective_date on this branch.

A couple of follow-ups before merge:

1. Tests
These new tests are currently failing; please fix them:

- TestGetQueuedTransactionsForCoalescing_PreservesEffectiveDate
- TestGetQueuedTransactionsForSourceCoalescing_PreservesEffectiveDate
- TestGetQueuedTransactionsForDestinationCoalescing_PreservesEffectiveDate
- TestGetStuckQueuedTransactions_PreservesEffectiveDate
(Looks like a time.Time vs *time.Time assertion mismatch, and a "USDs" typo)

2. GET by id and reference
GET /transactions/:id and GET /transactions/reference/:reference still omit effective_date, even when the field is present in the DB / filter response. Could you include it in those read paths as well?

TheFifthRider:
> Thanks @TheFifthRider for this PR — the coalescing/stuck-queue fix looks good, I was able to reproduce the bug and confirm the APPLIED rows keep effective_date on this branch.
> 
> A couple of follow-ups before merge:
> 
> 1. Tests
>    These new tests are currently failing; please fix them:
> 
> * TestGetQueuedTransactionsForCoalescing_PreservesEffectiveDate
> * TestGetQueuedTransactionsForSourceCoalescing_PreservesEffectiveDate
> * TestGetQueuedTransactionsForDestinationCoalescing_PreservesEffectiveDate
> * TestGetStuckQueuedTransactions_PreservesEffectiveDate
>   (Looks like a time.Time vs *time.Time assertion mismatch, and a "USDs" typo)
> 
> 2. GET by id and reference
>    GET /transactions/:id and GET /transactions/reference/:reference still omit effective_date, even when the field is present in the DB / filter response. Could you include it in those read paths as well?

Thanks for the review @Gee-Hive! I believe I've fixed the tests and covered the additional cases. Let me know if there's any other changes you'd like me to make!

Gee-Hive (commented):
Thanks @TheFifthRider  — the follow-ups look good. 
GET /transactions/:id and GET /transactions/reference/:reference now return effective_date, and the earlier test failures look fixed.

A few leftovers, none of them blocking the original queued though:

A few other read paths still drop effective_date, especially paginated list (GetTransactionsPaginated). Criteria search, grouped transactions, and shadow lookups have the same gap. Those won’t recreate the backdate bug, but the field will still be missing on those responses.

The new tests only check that the SQL scan fills the field. They don’t run the real apply. Worth adding more coverage for:

- coalescing and stuck-queue recovery, where the original date must land on the APPLIED child
- a skip_queue: false create → apply → GET
- paginated retrieval
- both a set effective_date and a null one (the coalescing tests already insert a null row, but never assert it stays nil)
- Prefer effectiveDate.Equal(...) over assert.Equal on *time.Time — same pattern as GetTransactionsByParent.

Gee-Hive (approved):

PR #388, opened 2026-09-04
feat(monitors): edge-triggered balance monitors, on by default

### LINKED ISSUE #318: Balance monitors are level-triggered (fire on every transaction while condition holds), not edge-triggered — no way to alert once on threshold crossing
## Summary

Balance monitors (`POST /balance-monitors`) re-fire the `balance.monitor` webhook on **every transaction that touches the balance while the condition evaluates true**, rather than once when the balance *crosses* the threshold. There is no edge-trigger, no dedupe, and no per-monitor "already fired" state. This makes monitors hard to use for the most common alerting case — "notify me when the balance drops below X" — because a balance sitting below the threshold produces a webhook on each subsequent debit/credit.

## Where this happens (verified against `main`, commit `f7a6a288fb1e0fcc2adcbedc76f7ed24b09c2c0c`)

**1. The condition check is purely point-in-time** — it only sees the *current* balance, never the previous value:

`model/model.go` (`CheckCondition`, ~L488–506):
```go
func (bm *BalanceMonitor) CheckCondition(b *Balance) bool {
    switch bm.Condition.Field {
    case "debit_balance":
        return compare(b.DebitBalance, bm.Condition.Operator, bm.Condition.PreciseValue)
    ...
    }
}
```

`BalanceMonitor` / `AlertCondition` (`model/balance.go`) carry no "last fired" / "was above" / cooldown field — stateless by construction.

**2. The check loop fires whenever the condition is currently true** — no transition gate:

`balance.go` (`checkBalanceMonitors`, ~L60–87):
```go
for _, monitor := range monitors {
    if monitor.CheckCondition(updatedBalance) {        // <- only gate: current-state truth
        go func(monitor model.BalanceMonitor) {
            err := l.SendWebhook(NewWebhook{
                Event:   "balance.monitor",
                Payload: monitor,
            })
        }(monitor)
    }
}
```

**3. It runs after every transaction touching the balance:**

`transaction_execution.go` (`runTransactionPostCommitWorkWithHooks`, ~L565–577):
```go
for _, balance := range orderedBalances {
    balanceMonitorSem <- struct{}{}
    go func() {
        defer func() { <-balanceMonitorSem }()
        l.checkBalanceMonitors(monitorCtx, balance)    // unconditional post-commit work
    }()
}
```

## Why this is a problem

For a `balance < threshold` monitor, the result is a burst of duplicate `balance.monitor` webhooks: one per debit while the wallet stays below the threshold, instead of a single alert on the down-crossing. Consumers must implement their own crossing-state machine (store the previous polarity/level and suppress repeats) to get usable alerts — which defeats much of the convenience of having a native monitor.

## Expected / desired behaviour

An option for **edge-triggered** monitors: fire `balance.monitor` once when the condition transitions `false → true`, and not again until it has recovered (`true → false`) and crossed back. Concretely, one of:

- A `trigger: "edge" | "level"` (or `"once" | "always"`) field on the monitor, defaulting to `level` for backward-compat; or
- Per-monitor last-evaluated state so the webhook is only sent on a state change.

## Workaround (current)

Poll the balance after each transaction and maintain an external state row (previous value / previous polarity), emitting your own event only on the transition. This is what we ended up doing for a low-balance notification.

## Environment

- Repo: `blnkfinance/blnk`, branch `main` @ `f7a6a288fb1e0fcc2adcbedc76f7ed24b09c2c0c`
- Component: balance monitors / webhooks

Shivam8584:
Closes #318.

> Builds on #391, which should merge first. Until it does, this diff includes its three commits. The feature is the last two.

Balance monitors fire on every committed balance update while their condition holds. A wallet below its low-balance threshold sends a webhook on every debit instead of one alert when it crosses.

Monitors now take a `trigger`:

- `edge` fires once when the condition goes from false to true, then stays quiet until it has been false again.
- `level` keeps the current behaviour.

As agreed on #318, `edge` is the default.

```jsonc
POST /balance-monitors
{
  "balance_id": "bln_...",
  "trigger": "edge",  // optional, defaults to "edge"
  "condition": { "field": "balance", "operator": "<", "value": 100, "precision": 100 }
}
```

The read endpoints return `trigger` and `condition_state`.

## How it works

Monitor checks run in goroutines after the balance lock is released, so evaluations of one monitor can run at the same time and arrive out of order. The state change is therefore a single conditional `UPDATE`, and only the evaluation that changes the row sends the webhook:

```sql
UPDATE blnk.balance_monitors
   SET condition_state = $2, state_version = $3, state_changed_at = NOW()
 WHERE monitor_id = $1
   AND balance_id = $4                      -- skip evaluations for a balance the monitor has left
   AND condition_state IS DISTINCT FROM $2  -- one concurrent evaluation wins
   AND state_version < $3                   -- skip evaluations older than the last one applied
```

`$3` is the balance's `version`, which only increases after a successful write. The `balance_id` check matters because versions are counted per balance. Without it, a stale evaluation from a monitor's previous balance could store a version the new balance won't reach for a long time, and the monitor would go quiet.

If the webhook can't be enqueued, the transition is rolled back so the next transaction can retry the same crossing.

Level monitors do no extra database work. Edge monitors add one indexed single-row update per monitor per committed balance update, outside the lock. Against a real Postgres that costs 0.13 ms when nothing changes and 0.89 ms on a crossing.

## Upgrade behaviour needs your call

Existing monitors migrate as `edge` with `condition_state = false`. A monitor whose condition already holds fires once on its balance's next transaction, then goes quiet. An install with many wallets already past their threshold gets one alert per monitor at upgrade.

The alternative is to migrate existing monitors as `level` and default only new ones to `edge`. That avoids the one-time burst, but old and new monitors would then behave differently with nothing in the API to tell them apart. I went with the burst and am happy to switch.

## Testing

- **Concurrency:** 24 goroutines race one crossing against real Postgres, and exactly one sends.
- **Property tests:** random balance walks where the expected alert count is computed independently of the implementation. Edge never sends more than level.
- **Coverage:** every condition field and operator, inflight commit and void, the batch path, and moved, duplicate and deleted monitors.
- **Guards:** each check in the `UPDATE` was removed in turn, and a test failed every time.
- **Live run** with `blnk start` and `blnk workers`: the same ledger activity delivered 2 webhooks from an edge monitor and 3 from a level one.

`trigger` will need a line in `blnkfinance/docs`. I'll open that once the API shape is settled.

### DIFF
--- a/api/balance.go
+++ b/api/balance.go
@@ -321,8 +321,18 @@ func (a Api) UpdateBalanceMonitor(c *gin.Context) {
 		return
 	}
 
+	// This handler binds the internal model rather than a request struct, so the
+	// trigger is resolved here or the update path becomes the one way to put an
+	// unvalidated value in front of the table's check constraint.
+	trigger, err := model.NormalizeTrigger(monitor.Trigger)
+	if err != nil {
+		respondCode(c, apierror.ErrGenValidation, err.Error(), nil, withLegacyKey("errors"))
+		return
+	}
+	monitor.Trigger = trigger
+
 	monitor.MonitorID = id
-	err := a.blnk.UpdateMonitor(c.Request.Context(), &monitor)
+	err = a.blnk.UpdateMonitor(c.Request.Context(), &monitor)
 	if err != nil {
 		respondError(c, err, withUpgrade(apierror.ErrGenNotFound, apierror.ErrBalMonitorNotFound))
 		return
--- a/api/balance_monitor_concurrency_test.go
+++ b/api/balance_monitor_concurrency_test.go
@@ -1,0 +1,203 @@
+/*
+Copyright 2024 Blnk Finance Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"sync"
+	"testing"
+
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	blnkservice "github.com/blnkfinance/blnk"
+	model2 "github.com/blnkfinance/blnk/api/model"
+	"github.com/blnkfinance/blnk/internal/request"
+	"github.com/blnkfinance/blnk/model"
+)
+
+// TestBalanceMonitorEdge_ConcurrentTransactions is the whole point of the
+// design, exercised through the product rather than the datasource: monitor
+// checks are dispatched into detached goroutines once the balance lock is
+// released, so several of them evaluate the same crossing at once. The consumer
+// must still be told once.
+func TestBalanceMonitorEdge_ConcurrentTransactions(t *testing.T) {
+	e := setupMonitorE2E(t)
+
+	funding := e.newBalance(t)
+	wallet := e.newBalance(t)
+	monitor := e.createMonitor(t, wallet.BalanceID, model.TriggerEdge, 500)
+
+	const senders = 12
+
+	var wg sync.WaitGroup
+	for i := 0; i < senders; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			payloadBytes, _ := request.ToJsonReq(&model2.RecordTransaction{
+				Amount:         200,
+				Precision:      1,
+				Reference:      gofakeit.UUID(),
+				Description:    "concurrent monitor crossing",
+				Currency:       "USD",
+				Source:         funding.BalanceID,
+				Destination:    wallet.BalanceID,
+				SkipQueue:      true,
+				AllowOverDraft: true,
+			})
+			resp, _ := SetUpTestRequest(TestRequest{
+				Payload: payloadBytes,
+				Method:  "POST",
+				Route:   "/transactions",
+				Router:  e.router,
+			})
+			assert.Equal(t, http.StatusCreated, resp.Code)
+		}()
+	}
+	wg.Wait()
+
+	// Every one of those transactions left the wallet past the threshold, so
+	// most of the evaluations saw a true condition.
+	e.awaitWebhooks(t, 1)
+	e.settle()
+	assert.Equal(t, 1, e.webhooks(), "%d concurrent transactions past the threshold must produce one alert", senders)
+	assert.True(t, e.monitorState(t, monitor.MonitorID))
+}
+
+// TestBalanceMonitorEdge_MixedTriggersOnOneBalance pins that the two modes stay
+// independent when they watch the same balance.
+func TestBalanceMonitorEdge_MixedTriggersOnOneBalance(t *testing.T) {
+	e := setupMonitorE2E(t)
+
+	funding := e.newBalance(t)
+	wallet := e.newBalance(t)
+	edge := e.createMonitor(t, wallet.BalanceID, model.TriggerEdge, 500)
+	level := e.createMonitor(t, wallet.BalanceID, model.TriggerLevel, 500)
+
+	e.transfer(t, funding.BalanceID, wallet.BalanceID, 600) // crossing: edge + level
+	e.awaitWebhooks(t, 2)
+
+	e.transfer(t, funding.BalanceID, wallet.BalanceID, 100) // still past: level only
+	e.awaitWebhooks(t, 3)
+
+	e.transfer(t, funding.BalanceID, wallet.BalanceID, 100) // still past: level only
+	e.awaitWebhooks(t, 4)
+
+	// The totals above could be satisfied by the wrong split, so pin each
+	// monitor's own count: that is what "independent" means here.
+	assert.Equal(t, 1, e.webhooksFor(edge.MonitorID), "the edge monitor owes one alert, on the crossing")
+	assert.Equal(t, 3, e.webhooksFor(level.MonitorID), "the level monitor owes one per update while the condition holds")
+
+	assert.True(t, e.monitorState(t, edge.MonitorID))
+	assert.False(t, e.monitorState(t, level.MonitorID), "a level monitor keeps no state")
+}
+
+// TestBalanceMonitorEdge_InflightCommit pins that the inflight settlement path
+// reaches the monitors too. It arrives by way of RecordTransaction, so it shares
+// the post-commit work, but it is the path a real wallet balance actually moves
+// on and is worth holding still.
+func TestBalanceMonitorEdge_InflightCommit(t *testing.T) {
+	e := setupMonitorE2E(t)
+
+	funding := e.newBalance(t)
+	wallet := e.newBalance(t)
+	monitor := e.createMonitor(t, wallet.BalanceID, model.TriggerEdge, 500)
+
+	payloadBytes, _ := request.ToJsonReq(&model2.RecordTransaction{
+		Amount:         900,
+		Precision:      1,
+		Reference:      gofakeit.UUID(),
+		Description:    "inflight crossing",
+		Currency:       "USD",
+		Source:         funding.BalanceID,
+		Destination:    wallet.BalanceID,
+		Inflight:       true,
+		SkipQueue:      true,
+		AllowOverDraft: true,
+	})
+	var queued model.Transaction
+	resp, _ := SetUpTestRequest(TestRequest{
+		Payload:  payloadBytes,
+		Response: &queued,
+		Method:   "POST",
+		Route:    "/transactions",
+		Router:   e.router,
+	})
+	require.Equal(t, http.StatusCreated, resp.Code)
+	require.NotEmpty(t, queued.TransactionID)
+
+	// Inflight money is not the balance, so nothing has crossed yet.
+	e.settle()
+	assert.Equal(t, 0, e.webhooks(), "an inflight transaction has not moved the balance")
+
+	commitBytes, _ := request.ToJsonReq(&model2.InflightUpdate{Status: blnkservice.InflightActionCommit, SkipQueue: true})
+	commitResp, _ := SetUpTestRequest(TestRequest{
+		Payload: commitBytes,
+		Method:  "PUT",
+		Route:   fmt.Sprintf("/transactions/inflight/%s", queued.TransactionID),
+		Router:  e.router,
+	})
+	require.Equal(t, http.StatusOK, commitResp.Code)
+
+	e.awaitWebhooks(t, 1)
+	assert.True(t, e.monitorState(t, monitor.MonitorID))
+}
+
+// TestBalanceMonitorEdge_WebhookPayload checks what a consumer actually
+// receives, not just that something was queued.
+func TestBalanceMonitorEdge_WebhookPayload(t *testing.T) {
+	e := setupMonitorE2E(t)
+
+	funding := e.newBalance(t)
+	wallet := e.newBalance(t)
+	monitor := e.createMonitor(t, wallet.BalanceID, model.TriggerEdge, 500)
+
+	e.transfer(t, funding.BalanceID, wallet.BalanceID, 900)
+	e.awaitWebhooks(t, 1)
+
+	tasks := e.pendingTasks()
+
+	// NewWebhook serialises its payload under "data", not "payload".
+	var payload struct {
+		Event string               `json:"event"`
+		Data  model.BalanceMonitor `json:"data"`
+	}
+	found := false
+	for _, task := range tasks {
+		var probe struct {
+			Event string `json:"event"`
+		}
+		if err := json.Unmarshal(task.Payload, &probe); err != nil || probe.Event != "balance.monitor" {
+			continue
+		}
+		require.NoError(t, json.Unmarshal(task.Payload, &payload))
+		found = true
+		break
+	}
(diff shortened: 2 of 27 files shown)


Shivam8584:
## Why the transition is one statement, and why each guard is load-bearing

Monitor checks are dispatched into detached goroutines once the balance lock is released:

```go
balanceMonitorSem <- struct{}{}
go func() {
    defer func() { <-balanceMonitorSem }()
    l.checkBalanceMonitors(monitorCtx, balance)
}()
```

So several evaluations of the same monitor can be in flight at once, in any order. Each guard was removed in turn to confirm it earns its place.

### `condition_state IS DISTINCT FROM $2`

Makes the read and the write one statement, so concurrent evaluations serialise on the row and only one can observe the transition.

Removed, with 24 goroutines racing one crossing:

```
--- FAIL: TestTransitionMonitorState_ConcurrentEvaluations_RealDB
    Error: Not equal:  expected: 1
                       actual  : 5
```

Five of twenty-four each believed they owned the crossing — five duplicate alerts, which is worse than the level-triggered burst this issue is about, because it is nondeterministic.

### `state_version < $3`

The goroutines are unordered, so an evaluation of balance 90 can land *after* an evaluation of balance 110. Without the fence it flips the state back to triggered, alerts on a balance that has already recovered, and then swallows the next genuine crossing.

`Balance.Version` is a sound fencing token: it is incremented only after a successful persist (`database/balance.go`), under the optimistic lock, so it is strictly increasing per balance and unique per committed update. It also makes the design idempotent when the same balance appears twice in one post-commit batch — the second call sees `$3 < $3` and matches nothing.

Removed: `TestTransitionMonitorState_VersionFence_RealDB` fails. A test drives 99 stale evaluations against a triggered monitor and none of them may disarm it.

### `balance_id = $4`

This one I did not have in the original design; it came out of hunting after the fact, and it is the one that would have bitten in production.

Versions count **per balance**, and the monitor list is cached **per balance**. For a few minutes after a monitor is moved to a different balance, a node can still evaluate it against the balance it used to watch. That stale evaluation writes `state_version` from the *old* balance's counter — which can be thousands ahead of the new balance's — and every subsequent genuine crossing is fenced out. The monitor goes silent, permanently in practice.

Matching `balance_id` makes the state write self-validating: an evaluation for a balance this monitor no longer watches matches no rows and touches nothing.

Removed:

```
--- FAIL: TestTransitionMonitorState_IgnoresAnotherBalance_RealDB
    Error: Should be true
    Messages: a real crossing on the monitor's own balance must not be fenced out
              by an evaluation for a different balance
```

Note the asymmetry this exposes: under level triggering a stale cached monitor only causes a spurious webhook, but under edge triggering it *loses* alerts. That is why the cache invalidation fix (`UpdateMonitor` now invalidates the balance a monitor moved away from, not just the one it moved to) had to land alongside it.

### Delivery semantics

An edge fires once, so a transition this process owned but could not hand off is released:

```go
if err := l.sendMonitorWebhook(monitor); err != nil {
    notification.NotifyError(err)
    _ = l.datasource.ReleaseMonitorState(ctx, monitor.MonitorID, updatedBalance.BalanceID, updatedBalance.Version)
}
```

`ReleaseMonitorState` is fenced on version **equality**, not `<`, so it undoes our own write and nobody else's; once a newer evaluation has moved the monitor on, the release correctly does nothing.

This is deliberately at-most-once with a rollback, not at-least-once: the window is between two adjacent statements, and duplicates are the thing the feature exists to remove.

Shivam8584:
## The six pre-existing defects, with the evidence

These were found while mapping the monitor surface for this change. Each was reproduced first, then fixed, then re-validated by reverting the fix and confirming its test fails again. They are in commits `e5380ca` and `1c4c8a7`, separate from the feature.

### 1. A legacy row silenced every monitor on its balance

`precision` and `precise_value` were added as **nullable** columns (`sql/1715045770.sql`), so every monitor created before that migration carries NULLs — and both were scanned into plain Go numbers.

```
GetBalanceMonitors with a NULL precise_value -> monitors=0 err=INTERNAL_SERVER_ERROR: Failed to scan monitor data
GetMonitorByID     with a NULL precise_value -> monitor=false err=INTERNAL_SERVER_ERROR: Failed to retrieve monitor
```

The blast radius is the bad part: `checkBalanceMonitors` gives up when the fetch errors, so **one** unreadable row disabled **every** monitor on that balance, silently. Both columns are read through `COALESCE(..., 0)` now, matching what `CreateMonitor` already defaults them to.

### 2. Equality monitors could never fire

The check constraint stores equality as `'='` and rejects `'=='`; `compare()` only handled `'=='`. So `balance = 100` was accepted, stored, and silently never fired — equality monitors were entirely unusable.

```
CreateMonitor with operator '=' -> err=<nil>
stored operator="="; CheckCondition on an exactly-equal balance -> false
```

`compare()` accepts both spellings now, and `ValidateMonitorCondition` constrains the operator to the set the ledger can actually evaluate, so a typo is a 400 rather than a monitor that never fires.

Worth flagging on upgrade: monitors already stored with `=` will start firing. That is what their author asked for, but it is a behaviour change.

### 3. Every update cleared the callback URL

`BalanceMonitor.CallBackURL` is tagged `json:"-"`, so a request can never populate it — but the UPDATE wrote it anyway, meaning every `PUT` replaced the stored URL with an empty string.

```
call_back_url before update="https://example.com/hook" after update=""
```

The update no longer touches the column. (Making the URL settable is a separate question; this just stops the data loss.)

### 4. A threshold of zero was impossible

`validation.Required` treats a zero number as absent, so **`balance > 0`, `balance != 0` and `balance <= 0` were all rejected with a 400** — arguably the most natural monitors there are. Found because a matrix test on `!=` failed at *creation*, not on the alert logic.

`Precision` keeps the rule, because a zero multiplier is meaningless. A zero threshold is not.

### 5. A fractional threshold returned 500

`AlertCondition.Value` is a `float64` and the API accepts one, but the column was a `BIGINT`, so `balance < 100.50` failed at the database and surfaced as an internal error.

```
CreateMonitor with Value=100.5 -> err=INTERNAL_SERVER_ERROR: Database error occurred
```

Widened to `DOUBLE PRECISION`. `precise_value` is what the ledger actually compares against, so this changes no evaluation — it only stops rejecting thresholds the API already advertises.

### 6. The list endpoint disagreed with the detail endpoint

`GetAllMonitors` did not select precision or precise_value:

```
GetMonitorByID -> precision=100 precise_value=10000
GetAllMonitors -> precision=0   precise_value=<nil>
```

Two smaller corrections to the create payload came with it: `meta_data` was accepted, validated and dropped (there is no column for it), and `description` is a real column that only an update could set.

### Unrelated, but found by the same sweep

`internal/cache`'s tests configured the package with an empty `Configuration`. `config.MockConfig` validates before it stores and **returns silently when validation fails**, so those calls did nothing — the tests only worked because `TestSet` ran first and left a usable config in the global store. Under `-shuffle` they get a nil cache and panic. Fixed in `da0bd7a`; 15/15 shuffled runs pass, where most failed before. Happy to pull that commit out if you'd rather it went separately.

Shivam8584:
## How this was verified

### Layers

| layer | what it proves |
| --- | --- |
| **Concurrency, real Postgres** | 24 goroutines race one crossing; exactly one owns it. |
| **Property-based** | Randomised ledger walks where the expected alert count is computed from the balance series *independently of the implementation*. Five seeds × 14 steps, edge and level, plus the property the change exists to deliver: edge is never noisier than level. |
| **Coverage matrix** | All six condition fields, each driven by the ledger movement that actually changes it; all six operators; the coalescing batch path (a second post-commit caller); inflight commit **and** void; negative thresholds; twelve monitors on one balance; duplicate monitors; a monitor created while its condition already holds; deletion while triggered; all three read endpoints. |
| **End-to-end per defect** | Each fix re-checked at the level a user meets it — a legacy row no longer silences its siblings, a moved monitor follows its balance, fractional and zero thresholds work through the API, list and detail agree. |
| **Live deployment** | Real `blnk start`, real `blnk workers`, real HTTP receiver. |

### The inflight cases are worth calling out

An inflight transaction does **not** move `balance`, so a monitor on `balance` stays quiet until the commit — and a monitor on `inflight_credit_balance` fires on the *inflight* and re-arms on the commit (which drains it) or on the void (which releases it). All four are pinned.

### Mutation testing

Tests that cannot fail prove nothing, so every guard and every fix was reverted in turn:

```
edge state guard (concurrent duplicates)       bug reproduced: CAUGHT   with fix: PASS
version fence (stale evaluation)               bug reproduced: CAUGHT   with fix: PASS
balance guard (monitor moved)                  bug reproduced: CAUGHT   with fix: PASS
unset trigger defaults to edge                 bug reproduced: CAUGHT   with fix: PASS
NULL precision tolerated (legacy rows)         bug reproduced: CAUGHT   with fix: PASS
GetAllMonitors carries precision               bug reproduced: CAUGHT   with fix: PASS
update preserves the callback URL              bug reproduced: CAUGHT   with fix: PASS
equality operator fires                        bug reproduced: CAUGHT   with fix: PASS
zero threshold accepted                        bug reproduced: CAUGHT   with fix: PASS
cache invalidated on both balances             bug reproduced: CAUGHT   with fix: PASS
```

Making the edge path behave like level fails 3 of the 5 property seeds (the other two have `edges == levels` for their walk, so they cannot distinguish — which is itself a useful check on the seeds).

### Live run

Identical ledger activity against two wallets, webhooks actually **delivered** over HTTP by the worker, not merely enqueued:

```
== webhooks delivered per monitor, identical ledger activity ==
edge  (mon_105272f0-...): 2
level (mon_33e3584f-...): 3
```

Five transactions, two crossings each. The monitor was created with **no** `trigger` in the request body and came back `"trigger":"edge","condition_state":false`, so the new default is exercised over the wire. The delivered payload:

```json
{"event":"balance.monitor","data":{"balance_id":"bln_...","condition":{...},
 "condition_state":true,"monitor_id":"mon_...","trigger":"edge"}}
```

### Suite

24/24 packages green on plain, `-shuffle=on`, `-race` and `-count=3`, each on a freshly reset database. Zero data races, no retries. `golangci-lint run ./...` reports no issues; `go mod tidy` produces no churn.

### Two things worth knowing if you run this locally

`./cmd`'s `TestGetOrCreateHeartbeatIDAt` needs **cgo** — the heartbeat store uses `mattn/go-sqlite3`. With `CGO_ENABLED=0` the driver never registers and the test fails misleadingly. It is not a real failure; CI builds with cgo on.

`POST /transactions/recover` recovers *every* stale queued row in the database, so `TestRecoverQueuedTransactions` starts failing once earlier interrupted runs have left junk behind — on stock `main` too. Reset the database between full runs rather than reading it as a regression.

Shivam8584:
## Schema, compatibility, and what a reviewer should look at first

### Two migrations

`sql/1788542465.sql` — the feature:

```sql
ALTER TABLE blnk.balance_monitors ADD COLUMN trigger_type TEXT NOT NULL DEFAULT 'edge';
ALTER TABLE blnk.balance_monitors ADD CONSTRAINT balance_monitors_trigger_type_check
  CHECK (trigger_type IN ('edge', 'level'));
ALTER TABLE blnk.balance_monitors ADD COLUMN condition_state BOOLEAN NOT NULL DEFAULT FALSE;
ALTER TABLE blnk.balance_monitors ADD COLUMN state_version BIGINT NOT NULL DEFAULT 0;
ALTER TABLE blnk.balance_monitors ADD COLUMN state_changed_at TIMESTAMP;
```

`sql/1788542466.sql` — widens `value` from `BIGINT` to `DOUBLE PRECISION` (defect 5).

Both round-trip: `migrate down` rolls back all 39 cleanly and `migrate up` restores the schema identically.

The column is `trigger_type`, not `trigger`. `TRIGGER` is non-reserved in Postgres and would work, but every query in `database/balance.go` is hand-written and one unquoted keyword in a column list is a bad afternoon.

### Compatibility

- **Wire format is unchanged for existing clients.** `trigger` is optional; omitting it yields `edge`. Responses gain `trigger` and `condition_state`.
- `meta_data` on the create payload is removed from the struct — it was already being dropped, and Go ignores unknown JSON fields, so a client still sending it sees no change.
- **The one real behaviour change** is the default itself, plus the one-time flush for already-satisfied monitors described in the PR body.
- Monitors stored with operator `=` will begin firing (defect 2). Previously they never could.

### Lifecycle rules, made explicit

| event | rule | why |
| --- | --- | --- |
| **Create** | Starts armed. | A monitor created while the condition already holds fires on the balance's next transaction. Evaluating at creation would mean loading the balance and writing twice for a one-transaction difference. |
| **Update** | Always re-arms, and invalidates the cache on **both** the old and new balance. | The condition has been replaced, so state carried against the old one would swallow the first crossing of the new threshold. |
| **Read** | `condition_state` is returned; `state_version` is not. | The version is an internal fencing token; exposing it would only invite dependence on it. |
| **Evaluate** | The cached monitor supplies `trigger_type` and the condition only. | Enforced by shape: `TransitionMonitorState` takes IDs, so there is no cached state object to accidentally trust. |
| **Rolling upgrade** | An un-upgraded node keeps level-firing for up to 5 minutes; pre-upgrade cache entries decode with an empty trigger. | Harmless and self-healing — `model.NormalizeTrigger` is the single place that resolves an unset trigger to the default, and it is exercised against real Redis with the real codec. |

### Where I'd look first as a reviewer

1. `database/balance.go` — `TransitionMonitorState` / `ReleaseMonitorState`. The three guards are the whole design.
2. `balance.go` — `checkBalanceMonitors`. The level branch must remain byte-for-byte the old behaviour; the edge branch is ~25 lines.
3. `model/balance.go` — `NormalizeTrigger` is deliberately the *only* place that knows the default, so the API, the datasource and the read path cannot drift apart.

### Happy to change

- Split the feature and the fixes into separate PRs — the commits are already clean for it.
- Drop the `internal/cache` test commit if you'd rather it went on its own.
- Flip the upgrade behaviour to backfill existing rows as `level`, if you'd prefer no flush.

Shivam8584:
@jerry-enebeli @Gee-Hive this is ready for review when you have time. I've split it in two: #391 has seven pre-existing monitor bugs and should merge first, and this PR is now just the edge-trigger feature on top of it. There's one decision for you in the description, under "Upgrade behaviour".

